PR #13, opened 2022-11-17
Updates to Front-end

colinmfoster4723:
- Added Import Token button
- Added dynamic highlighting w/ tailwind
- Fixed small bugs


larrotiz (approved):
Thanks for all the help!

PR #14, opened 2022-11-22
try to fix tp issue

Allin-Moondao:
replace InjectedConnector with MetamaskConnector


larrotiz (approved):
Let's try it out, everything still seems to work on my end

PR #22, opened 2023-03-01
Ticket-0003: wagmi refresh issue

colinmfoster4723:
- could not find a way to reproduce the error
- use-preferred-network.ts was missing preferredNetwork dependency


colinmfoster4723:
Ticket 2 and 3 have been closed

PR #24, opened 2023-03-02
optimized shopify redirect

colinmfoster4723:


colinmfoster4723:
these updates have already been committed.

PR #145, opened 2024-07-29
Refactor

namedotget:
- Refactored chain default


pmoncada (approved):
Looks good to me. So do we now need to add useChainDefault() to every page going forward?

namedotget:
For now yes, when we move away from Polygon (Sweepstakes page is set to Polygon) then we can set the chain in _app.tsx and it will apply to every page.

PR #148, opened 2024-07-30
Refactor

namedotget:
- Adjusted links
- Adjusted images (replaced  w/ next/image)
- Refactor
- Adjusted useEffect deps to fix warnings in console

### DIFF
--- a/ui/components/ticket-to-space/SweepstakesWinners.tsx
+++ b/ui/components/ticket-to-space/SweepstakesWinners.tsx
@@ -68,40 +68,41 @@ function WinnerSkeleton() {
 export function SweepstakesWinners({ ttsContract, supply }: any) {
   const [winners, setWinners] = useState<Winner[]>([])
 
-  async function getWinners() {
-    const winners = []
-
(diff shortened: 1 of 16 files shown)


ui/components/ticket-to-space/SweepstakesWinners.tsx
```diff
@@ -68,40 +68,41 @@ function WinnerSkeleton() {
 export function SweepstakesWinners({ ttsContract, supply }: any) {
   const [winners, setWinners] = useState<Winner[]>([])
 
-  async function getWinners() {
```
pmoncada, line 71: Do we need any of this logic any more? Could we just get rid of this entirely?

pmoncada (approved):
It generally looks fine but I wonder how much of this we can just remove entirely to simplify the codebase.

ui/components/ticket-to-space/SweepstakesWinners.tsx
namedotget: I can archive it but I think we should keep the code.

PR #150, opened 2024-07-31
Adjusted team members

namedotget:
- Consolidated team member cards to one per address w/ multiple hats
- Removed unused props


pmoncada (approved):
Looks good to me but checking for each hat to see if the ID matches is an O(N^2) - it doesn't matter too much for lower sized teams. Using a map that keys on address and resolves to a set of hats would automatically deduplicate on insertion and make the operation O(N) for creating the set and then O(N) for reading through it.

Not needed to modify here but just a thought for the future. Unless if you're seeing a drop in performance a loading time

MitchMcQuinn (approved):

PR #159, opened 2024-08-08
Refactor

namedotget:
- Added passHref to links
- Swapped  for next/image
- Added useEffect deps
- Fixed warnings in console on build
- Removed old citizen nft from onboarding flow


ui/lib/tokens/hooks/useValidVP.tsx
```diff
@@ -10,36 +10,36 @@ export function useValidVP(address: string | undefined) {
...
+      const L1Lock = await L1VMooneyContract.call('locked', [address])
+      const L2Lock = await L2VMooneyContract.call('locked', [address])
+
+      if (L1Lock[0] > 0 || L2Lock[0] > 0) {
+        setValidLock(true)
+      } else {
+        setValidLock(false)
+      }
+
+      const totalMooneyLocked = L1Lock[0]
```
pmoncada, line 34: Ternary operators are really hard to parse and often prone to logical errors. Can you break this out into an if / else?

pmoncada (changes requested):

ui/lib/tokens/hooks/useValidVP.tsx
pmoncada: Also, you're checking for L1Lock[0] and L2Lock[0] repeatedly. You could just store those as variables so you don't have to index into it repeatedly.

namedotget: Can do. Once we merge entity-registration into main I will deprecate this hook as we're now fetching voting power from snapshot.

pmoncada (approved):

PR #161, opened 2024-08-09
Updated privy funding

namedotget:
- Updated privy dep and privy funding


ui/yarn.lock
pmoncada, line 1: Should we put yarn.lock etc in the git-ignore?

pmoncada (approved):

ui/yarn.lock
namedotget: yarn.lock is required to ensure that node_modules is the same on all machines.

PR #199, opened 2024-09-26
Updates

namedotget:
- Added stable token balances to dropdown if balance > 0
- Added stable token support to sending modal
- Adjusted get-mooney and lock descriptions, removed padding
- Added usd cost to swap component
- Removed unused code and imports


pmoncada (approved):
LGTM

PR #200, opened 2024-09-26
Added required chain for web3 buttons

namedotget:
- Adjusted styling of modal
- Added default chain const to config
- Added required chain prop on web3 button


pmoncada:
What's up with all the failing tests? Can we fix those or remove them if they're no longer useful?

pmoncada (approved):
Looks fine but we should definitely look into those automated tests that are failing

namedotget:
> What's up with all the failing tests? Can we fix those or remove them if they're no longer useful?

Fixed

PR #207, opened 2024-10-02
Added tests for layout components

namedotget:


ui/cypress.config.ts
```diff
@@ -14,12 +14,11 @@ export default defineConfig({
...
       bundler: 'webpack',
     },
-    // excludeSpecPattern: [
-    //   'cypress/integration/layout/**',
-    //   'cypress/integration/lock/**',
-    //   'cypress/integration/privy/**',
-    //   'cypress/integration/shopify/**',
-    //   'cypress/integration/directory/**',
-    // ],
+    excludeSpecPattern: [
```
pmoncada, line 17: Should we get rid of this commented out block? Or is the idea to revive these soon?

pmoncada (approved):

ui/cypress.config.ts
namedotget: I was using that to skip tests when developing, I recently figured out that I can specifiy a certain file rather than running all at once, I can remove it.

PR #208, opened 2024-10-03
Updates

namedotget:
- Added address component and added it to citizens and teams page
- Disabled extend subscription for citizens and teams


ui/pages/citizen/[tokenId].tsx
```diff
@@ -234,7 +235,7 @@ export default function CitizenDetailPage({
                         )}
                       </div>
                     ) : null}
-                    {address === nft.owner ? (
+                    {/* {address === nft.owner ? (
```
Comment (pmoncada) [inline ui/pages/citizen/[tokenId].tsx:238]: Remove commented out code

pmoncada (approved):

ui/pages/citizen/[tokenId].tsx
Comment (namedotget) [reply ui/pages/citizen/[tokenId].tsx:238]: That is for extending the subscription. We should keep that code so that we can easily re-enable it when we're ready.

PR #215, opened 2024-10-06
Allow citizen image generator to fallback to input image

namedotget:


ui/lib/utils/images.ts
```diff
@@ -0,0 +1,54 @@
+export async function fitImage(
+  file: File,
+  maxWidth: number,
+  maxHeight: number
+): Promise<File> {
+  return new Promise((resolve, reject) => {
+    const img = new Image()
+    img.onload = () => {
```
pmoncada, line 8: What happens if the image is not the right ratio of height and width? How does it get cropped then?

pmoncada: Should we do error handling if the image dimensions are really off? Have you tried that to see what happens?

Could be good to add some tests to this function to make sure it's working as expected

pmoncada (approved):

ui/lib/utils/images.ts
namedotget: I used this image to test it, it should fit any image into the specified maxWidth and maxHeight
gioele-muscolino-screenshot00000

PR #233, opened 2024-10-18
Updates

namedotget:
- Removed newsletter check from CitizenMetadataModal and typeform function
- Added netlify config to prevent caching
- Adjusted types of StandardButton and added a hoverEffect bool to turn off the hover effect
- Added marketplace linking
- Adjusted citizen check messages on job and team pages


jaderiverstokes (approved):

ryand2d (approved):

ui/pages/join-us.tsx
ryand2d, line 1: this og image should actually be for /join not join-us

ryand2d (approved):

ryand2d (approved):

PR #240, opened 2024-11-07
Bug Fixes

namedotget:
- Fixed issues with balances
- Small styling adjustments
- Removed unused files


pmoncada (approved):

pmoncada (approved):
It would be good to add a test here to make sure the wallet provider is working as expected? Not sure how you'd do that, but if possible it would be good to add something to test for this exact but and show that the test now passes

PR #243, opened 2024-11-07
Archived current projects

namedotget:
- Archived current projects page
- Removed current projects from nav


pmoncada (approved):

pmoncada:
This is definitely something we want to have soon - talk to mitchie about how we could automate this w/ governance to create a project table entry automatically from the proposal.

PR #244, opened 2024-11-08
Added marketplace listing non-citizen upcharge

namedotget:
- Marketplace listings cost 10% more for non-citizens
- Added token value truncation
- Rounded vmooney value on citizen profile


pmoncada (changes requested):
Two things here...

1.) We need to communicate when people are listing that non-citizens will get charged 10% more.
2.) Please add a test to demonstrate the new feature!

namedotget:
> Two things here...
> 
> 1.) We need to communicate when people are listing that non-citizens will get charged 10% more. 2.) Please add a test to demonstrate the new feature!

In the 2nd commit of this PR I added a message to the TeamMarketplaceListingModal

ui/lib/subscription/getMarketplaceListings.ts
```diff
@@ -0,0 +1,38 @@
+import { Arbitrum, Sepolia } from '@thirdweb-dev/chains'
```
pmoncada, line 1: Nice! Could you also add a test for the expiration? Add a listing w/ an expiration and make sure that it renders as expired.

ui/cypress/e2e/marketplace.cy.ts
```diff
@@ -0,0 +1,28 @@
+import getMarketplaceListings from '@/lib/subscription/getMarketplaceListings'
```
pmoncada, line 1: Could we also test the inverse? E.g. a signed in Citizen should see this without the upcharge

pmoncada (changes requested):

pmoncada (approved):
Nice! Awesome that the privy provider is working

PR #245, opened 2024-11-08
Added pretty links for citizens and teams

namedotget:


pmoncada (changes requested):
Nice - Looks great!

The only thingg that I'd add would be good to add some unit tests. In particular around checking that it generates a unique ID if there's a clash between pretty links / other potential corner cases with the pretty links

ui/pages/network.tsx
```diff
@@ -330,10 +337,25 @@ export async function getStaticProps() {
...
+  const prettyLinks = {
+    team: {},
+    citizen: {},
+  }
+  const teamPrettyLinkData = filteredValidTeams.map((nft: any) => ({
+    name: nft?.metadata?.name,
+    id: nft?.metadata?.id,
+  }))
+  const { idToPrettyLink: teamIdToPrettyLink } =
+    generatePrettyLinks(teamPrettyLinkData)
```
pmoncada, line 350: The way I'm understanding this we're regenerating all the links every time we load the page...

Would it make sense to just store the unique text ID for each in tableland? That way we're not doing an O(N) computation each time we render the page, and just doing it once when a team is generated?

Especially as we scale with more Citizens and Teams, it might make sense to just store this data v.s. generating  it on the fly every time. Also, this could cause some issues where if a team is deleted, then the "link" for a team further down the list might get changed (there's no longer a conflict), so it's better if these links are just static after being generated, versus dynamic and calculated each time.

pmoncada (changes requested):

ui/pages/network.tsx
namedotget: If the pretty link already exists then the tokenId is appended to the end of the pretty link

pmoncada (approved):
Looks good to me - it's too bad all the links will now have a number, but I think this method makes things more straightforward on our end.

PR #246, opened 2024-11-08
Improve profile editing

namedotget:
- Simplified profile editing by only asking for email in typeform and asking for the rest of info in a separate form that is prefilled with the citizen/team data


ui/components/subscription/CitizenMetadataModal.tsx
```diff
@@ -124,21 +197,54 @@ export function CitizenMetadataModal({ nft, selectedChain, setEnabled }: any) {
         )}
         {stage === 1 && (
           <>
-            <Widget
-              className="w-[100%] md:w-[100%]"
-              id={process.env.NEXT_PUBLIC_TYPEFORM_CITIZEN_FORM_ID as string}
-              onSubmit={submitTypeform}
-              height={500}
-            />
+            <p>Would you like to update your email?</p>
```
pmoncada, line 200: Stages works here - but I wonder if you could just surface all of them togther at the same time?

So clicking the "Edit" icon would pop up a modal with three options:
1.) Edit Image
2.) Edit Email
3.) Edit Profile

And then clicking on one lets you go to it directly, versus having to click through stages each time?

I think that would be more intuitive, but it might be a pain to implement. If it ends up being a big restructure just ignore, but if it's easy to do then that could be better

ui/components/subscription/TeamMetadataModal.tsx
```diff
@@ -5,20 +5,71 @@ import { Widget } from '@typeform/embed-react'
...
+      <FormInput
+        label="Communications"
+        value={teamData.communications}
+        onChange={({ target }: any) =>
+          setTeamData((prev: any) => ({
+            ...prev,
+            communications: target.value,
+          }))
+        }
+        placeholder="Enter your communications link"
```
pmoncada, line 59: Communications Link is a little vague... Maybe Ryan has a better word for this.
Something like "Other socials Link"?

ui/components/subscription/TeamMetadataModal.tsx
```diff
@@ -5,20 +5,71 @@ import { Widget } from '@typeform/embed-react'
...
+function TeamMetadataForm({ teamData, setTeamData }: any) {
+  return (
+    <div className="w-full flex flex-col gap-2">
+      <FormInput
+        label="Name"
+        value={teamData.name}
+        onChange={({ target }: any) =>
+          setTeamData((prev: any) => ({ ...prev, name: target.value }))
+        }
+        placeholder="Enter your name"
```
pmoncada, line 32: Periods at the end of these statements would be good too

pmoncada (approved):

PR #270, opened 2024-12-10
Withdraw vMOONEY

jaderiverstokes:
Adds withdraw vMOONEY page

2 new contracts:

contracts/src/governance/VotingEscrowDepositor.sol
       - batches together transferring and locking of vMOONEY for a user
       // set the amounts of vMOONEY for each contributor at the end of the quarter
       increaseWithdrawAmounts(userAddresses, amounts)
       // allow a user to withdraw that vMOONEY, provided they have a lock and have approved the MOONEY to be transferred
       withdraw()

contracts/src/governance/SmartWalletChecker.sol
       - required by vMOONEY contract so that a VotingEscrowDepositor can call deposit_for


contracts/src/test/VotingEscrowDepositor.t.sol
```diff
@@ -0,0 +1,105 @@
...
+        // Mint tokens for the depositor contract to transfer
+        //token.mint(address(depositor), initialBalance);
+        vm.prank(user);
+        token.approve(address(escrowToken), depositAmount);
+        vm.prank(0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266);
+        depositor.sendVotingEscrowTokens(address(user), depositAmount);
+    }
+
+    function testTransferAndDepositFor() public {
+        //uint256 depositAmount = 500;
```
pmoncada, line 74: Can be deleted

contracts/src/governance/VotingEscrowDepositor.sol
```diff
@@ -0,0 +1,65 @@
...
+        }
+        delete withdrawAddresses;
+    }
+
+    function withdraw() external {
+        uint256 amount = availableToWithdraw[msg.sender];
+        require(amount > 0, "No amount available to withdraw");
+        availableToWithdraw[msg.sender] = 0;
+        require(token.transfer(msg.sender, amount), "Token transfer failed");
+        escrowToken.deposit_for(msg.sender, amount);
```
pmoncada, line 53: This assumes someone already has an existing lock right? Do we need to check that their existing lock already has 4 years?

ui/pages/withdraw.tsx
```diff
@@ -0,0 +1,107 @@
...
+import WebsiteHead from '../components/layout/Head'
+import StandardButton from '../components/layout/StandardButton'
+import Asset from '@/components/dashboard/treasury/balance/Asset'
+import { NoticeFooter } from '@/components/layout/NoticeFooter'
+import NetworkSelector from '@/components/thirdweb/NetworkSelector'
+
+export default function Withdraw() {
+  const userAddress = useAddress()
+  const router = useRouter()
+  const chain = process.env.NEXT_PUBLIC_CHAIN === 'mainnet' ? Arbitrum : Sepolia
```
namedotget, line 29: Instead of defining the chain this way try:
`const {selectedChain} = useContext(ChainContext);`

ui/pages/withdraw.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const { contract: votingEscrowDepositorContract } = useContract(
+    VOTING_ESCROW_DEPOSITOR_ADDRESSES[chain.slug]
```
namedotget, line 31: pretty sure adding the abi here speeds things up so thirdweb doesn't need to fetch it

namedotget:
Screenshot 2024-12-12 at 5 27 04 PM

contracts/src/governance/VotingEscrowDepositor.sol
jaderiverstokes: it does assume that. i added a check on frontend which will direct them to create a lock if they don't have one. if they run with no lock, the transaction will fail.

this contract doesn't do any checks for the lock duration.

pmoncada: Hmm, I guess it would be possible for someone to unfairly have a lower lock than someone else. Maybe we should enforce that at least their last lock date was within the last 6 months? Then there's some leeway with range, but it enforces that new locks are longer?

We'd have to tell people to use new wallets if they're contributing in new quarters? It's a little clunky.

Unless if you have a better solution? Or is there another way you had in mind of handling this situation where someone might have a 1 week lock vs 4 years?

pmoncada: Another alternative is that this is handled on the "front-end" and we ask every person listed on the project to lock for X amount of time before the retroactives run? And if they don't then they just aren't added to this contract?

What do you think?

jaderiverstokes: it's pretty easy to add some changes to the frontend, which could have people extend their lock to some minimum. it wouldn't be enforced by the withdraw contract tho. i think we will move to a deposit instead of withdraw flow in the future so this could be temporary for a quarter or two.

jaderiverstokes: discussed during standup, we will have Mitchie go through setting up the lock with each contributor. we will also include the frontend check for their lock time

jaderiverstokes: added frontend check for at least 6 months in lock

namedotget (approved):

PR #282, opened 2024-12-27
Bug fixes

namedotget:
- Added max width for profile names so that they don't get cutoff on larger screens
- Adjusted featured teams


jaderiverstokes (approved):
approved

PR #292, opened 2025-01-08
Projects

namedotget:
- Added /project page that displays all active projects
- Added /project/[id] page that displays the profile for a project


ui/pages/project/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+    })
+  }
+
+  const [tab, setTab] = useState<string>('active')
+  function loadByTab(tab: string) {
+    if (tab === 'active') {
+      setCachedNFTs(
+        input != '' ? filterBySearch(activeProjects) : activeProjects
+      )
+    } else if (tab === 'inactive') {
```
pmoncada, line 68: Maybe "past projects" rather than inactive?

pmoncada: Nothing we need to do now, but it would be cool to look at past quarters. Like looking on a quarter by quarter basis

pmoncada (approved):

ui/pages/project/index.tsx
namedotget: > Maybe "past projects" rather than inactive?

Screenshot 2025-01-07 at 8 39 46 PM

Thinking I'll keep the tab value the same but change the label to 'Past'

PR #299, opened 2025-01-10
fix bug where project wouldn't show up if id is 0, and don't allow gi…

jaderiverstokes:
…ving rewards to projects you contributed to


namedotget:
Screenshot 2025-01-10 at 11 40 04 AM

pmoncada (approved):

PR #349, opened 2025-01-24
Fix bugs encountered during Q4 2024 retroactives

jaderiverstokes:
- Take into account upfront payment
- Add in community rewards
- Account for users with different voting and citizen addresses
- Switch voting power hook to not require snapshot ID
- Check citizenship of all voters using SWR for performance
- Only allow votes on arbitrum


ui/components/nance/RetroactiveRewards.tsx
```diff
@@ -176,16 +168,36 @@ export function RetroactiveRewards({
     return userAddress && userVotingPower > 0
   }, [userVotingPower, userAddress])
 
-  // All projects need at least one citizen distribution to do iterative normalization
-  const isCitizens = useCitizens(chain, addresses)
-  const citizenDistributions = distributions?.filter((_, i) => isCitizens[i])
+  const isCitizenAddresses = useCitizens(chain, addresses)
+  const citizenVotingAddresses = [
+    '0x78176eaabcb3255e898079dc67428e15149cdc99', // payout for ryand2d.eth
```
pmoncada, line 173: Do we need to hardcode these? Is this just because their voting wallet is different?

pmoncada (approved):

ui/components/nance/RetroactiveRewards.tsx
jaderiverstokes: yeah these two seem to be the only which don't vote with their citizen account. not ideal but gets the job done

PR #352, opened 2025-01-28
Added create proposal button

namedotget:


pmoncada (commented):
Nice - while on this page could you also left align the proposals they match the header? The proposals are indented kind of weird and it looks a little "off" IMO

pmoncada (approved):

PR #385, opened 2025-02-19
User Feedback Improvements

namedotget:
- Fixed citizen and team links in nav
- Added list view for network
- Styling adjustments (globe icon, converted buttons to tabs)


pmoncada (changes requested):
While you're at it could you add a "Become a Citizen" "Create a Team" button at the top of the network, similar to how we do that on the proposals page with "Create a Proposal"

Also could you make the page moved centered with the list items? Small thing but on bigger screens it makes things look shifted strangely
image

Otherwise looks good to me!

pmoncada (approved):

PR #432, opened 2025-04-30
Add chainlink function to fetch totalSupply and balanceOf(address) across chain

jaderiverstokes:


fee-hook/script/base/Config.sol
```diff
@@ -1,51 +1,118 @@
...
 
     constructor() {
+        string memory ethJson = vm.readFile("../contracts/deployments/ethereum.json");
+        string memory arbJson = vm.readFile("../contracts/deployments/arbitrum.json");
+        string memory baseJson = vm.readFile("../contracts/deployments/base.json");
+        string memory sepJson = vm.readFile("../contracts/deployments/sepolia.json");
+        string memory polygonJson = vm.readFile("../contracts/deployments/polygon.json");
+        string memory arbSepJson = vm.readFile("../contracts/deployments/arbitrum-sepolia.json");
+
+        // vMOONEY doesn't exist on arbitrum-sepolia
```
pmoncada, line 38: lol this file is getting wild. love it. can't wait to make a beautiful diagram of the topology of all these contracts and how they interact

fee-hook/src/FeeHook.sol
```diff
@@ -161,48 +237,78 @@ contract FeeHook is BaseHook, OApp  {
...
+        bytes memory err
+    ) internal override {
+        if (err.length > 0) {
+            revert(string(err));  // bubble up the actual error
         }
-        uint256 totalSupply = IERC20(vMooneyAddress).totalSupply();
-        require(totalSupply > 0, "No vMooney supply");
-        uint256 userBalance = IERC20(vMooneyAddress).balanceOf(msg.sender);
+        // Update the contract's state variables with the response and any errors
+        (uint256 totalSupply, uint256 userBalance) = abi.decode(response, (uint256, uint256));
```
pmoncada, line 288: Okay... question...

Say I want to try and exploit this... could I increase my ownership right before withdrawing? Do we need to account for that? I guess they would be locking more MOONEY... but extracting current fees now, they would be increasing the total supply as well of vMOONEY right? Does the math math?

pmoncada: Do we need to calculate the amount of fees each person is getting at the time that it "enters" into the contract v.s. calculate it when people are withdrawing?

e.g.: 1 ETH is generated in fees at time X: Is the relevant proportion of fees each person gets calculated at time X, or when the withdraw event happens at time X + Y when the vMOONEY distributions may have changed?

fee-hook/script/base/Config.sol
jaderiverstokes: 😍

fee-hook/script/WithdrawFees.s.sol
```diff
@@ -0,0 +1,20 @@
+pragma solidity ^0.8.20;
+
+import "forge-std/console.sol";
+import "forge-std/Script.sol";
+import "../src/FeeHook.sol";
+import {Config} from "./base/Config.sol";
+
+contract MyScript is Script, Config {
```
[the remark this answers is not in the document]
jaderiverstokes: done

fee-hook/src/FeeHook.sol
jaderiverstokes: lets put this into a separate PR, this one is already quite large.

pmoncada (approved):

PR #444, opened 2025-05-08
Mission contribution and styling adjustments

namedotget:


pmoncada (approved):
add a test to make sure it works before merging!

PR #466, opened 2025-05-23
Adds cross chain payment for jb terminals

jaderiverstokes:
- Adds cross chain payment for jb terminals
- Switch deploy contract to create2 to allow the same address across chains
- Adds Miner so we can mine a salt for specific addresses, eg we can have all MoonDAO addresses staring with 0x.da0...

### DIFF
--- a/subscription-contracts/script/CrossChainPayConnect.s.sol
+++ b/subscription-contracts/script/CrossChainPayConnect.s.sol
@@ -1,0 +1,32 @@
+pragma solidity ^0.8.20;
+
+import "../src/CrossChainPay.sol";
+import "base/Config.sol";
+
+contract MyScript is Script, Config {
+    function addressToBytes32(address _addr) public pure returns (bytes32) {
+        return bytes32(uint256(uint160(_addr)));
+    }
+
+    function run() external {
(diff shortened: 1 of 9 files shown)


subscription-contracts/script/CrossChainPayConnect.s.sol
```diff
@@ -0,0 +1,32 @@
...
+    function run() external {
+        uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
+        vm.startBroadcast(deployerPrivateKey);
+
+        CrossChainPay pay = CrossChainPay(payable(CROSS_CHAIN_PAY_ADDRESS));
+        if(block.chainid == MAINNET) { //mainnet
+            pay.setPeer(LZ_EIDS[ARBITRUM], addressToBytes32(CROSS_CHAIN_PAY_ADDRESS));
+        } else if (block.chainid == ARBITRUM) { //arbitrum
+            pay.setPeer(LZ_EIDS[BASE], addressToBytes32(CROSS_CHAIN_PAY_ADDRESS));
+            pay.setPeer(LZ_EIDS[MAINNET], addressToBytes32(CROSS_CHAIN_PAY_ADDRESS));
```
namedotget, line 20: should arb have two peers? wondering why its different than the others

PR #475, opened 2025-05-28
Team adjustments

namedotget:
- Added hover effect for actions
- Hide jobs and marketplace sections if there are no jobs or listings
- Move treasury section to the top of the team profile

### DIFF
--- a/ui/pages/team/[tokenIdOrName].tsx
+++ b/ui/pages/team/[tokenIdOrName].tsx
@@ -523,6 +523,15 @@ export default function TeamDetailPage({
               )}
             </div>
           )}
+          {isManager && (
+            <TeamTreasury
+              multisigAddress={nft.owner}
+              multisigMooneyBalance={MOONEYBalance?.displayValue}
+              multisigNativeBalance={nativeBalance?.displayValue}
+              multisigDAIBalance={DAIBalance?.displayValue}
+              multisigUSDCBalance={USDCBalance?.displayValue}
+            />
+          )}
           {/* Header and socials */}
           <TeamMissions
             selectedChain={selectedChain}
@@ -606,7 +615,6 @@ export default function TeamDetailPage({
                   </SlidingCardMenu>
                 </div>
               </Frame>
-              {/* Jobs */}
               <TeamJobs
                 teamId={tokenId}
                 jobTableContract={jobTableContract}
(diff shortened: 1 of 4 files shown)


ui/pages/project/[tokenId].tsx
```diff
@@ -335,15 +340,15 @@ export default function ProjectProfile({
...
-            {isManager && (
-              <TeamTreasury
-                multisigAddress={owner}
-                multisigMooneyBalance={MOONEYBalance?.displayValue}
-                multisigNativeBalance={nativeBalance?.displayValue}
-                multisigDAIBalance={DAIBalance?.displayValue}
-                multisigUSDCBalance={USDCBalance?.displayValue}
-              />
-            )}
+            <TeamTreasury
```
Comment (pmoncada) [inline ui/pages/project/[tokenId].tsx:343]: If we're showing this to everyone we need to make sure the manage treasury buttons don't render on the page if they're not a manager

ui/pages/team/[tokenIdOrName].tsx
```diff
@@ -646,16 +645,16 @@ export default function TeamDetailPage({
...
-                  multisigNativeBalance={nativeBalance?.displayValue}
-                  multisigDAIBalance={DAIBalance?.displayValue}
-                  multisigUSDCBalance={USDCBalance?.displayValue}
-                />
-              )}
+
+              <TeamTreasury
+                isSigner={isSigner}
+                safeData={safeData}
+                multisigAddress={nft.owner}
```
Comment (pmoncada) [inline ui/pages/team/[tokenIdOrName].tsx:652]: It'd be good to only display ETH, and then display the rest if and only if they have that token in the treasury?

pmoncada (changes requested):

ui/pages/project/[tokenId].tsx
Comment (namedotget) [reply ui/pages/project/[tokenId].tsx:343]: https://github.com/Official-MoonDao/MoonDAO/pull/475/files#diff-744ec5bfb2eddb509467275c089f94b1e33666fd142bdd010213850c7870f897

ui/pages/team/[tokenIdOrName].tsx
Comment (namedotget) [reply ui/pages/team/[tokenIdOrName].tsx:652]: I'm pulling in all tokens via the safe api instead in a different PR that I will open here shortly

pmoncada (approved):

PR #476, opened 2025-05-28
Added linking for create mission flow

namedotget:
- /launch?status=create will now link directly to the create mission flow, prompting the user to login if they haven't already
- Adjusted linking of team requirement modal


ui/pages/launch.tsx
```diff
@@ -163,6 +168,27 @@ export default function Launch({ missions }: any) {
...
+  useEffect(() => {
+    if (status) {
+      shallowQuery({
+        status: status,
+      })
+    }
+  }, [status])
+
+  useEffect(() => {
+    if (router.query.status) {
```
pmoncada, line 180: Was the issue before that it was letting you go to the create flow if you weren't a team yet?

pmoncada (approved):

ui/pages/launch.tsx
namedotget: No? Just added a query param so that we can link directly to the create mission flow.

PR #482, opened 2025-05-29
Team Page, Team Treasury and Safe Upgrade

namedotget:
- Add hover effect to actions
- Hide jobs and marketplace if there are no jobs or listings
- Fetch assets from safe api for team treasury
- Upgrade safe deps and hooks


ui/const/icons.ts
```diff
@@ -0,0 +1,13 @@
+export const COIN_ICONS = {
+  ADA: '/coins/ADA.svg',
```
pmoncada, line 2: I don't forsee us using ADA or BNB or BTW (given we have WBTC) since those are native to other chains

pmoncada (approved):

PR #483, opened 2025-05-29
Safe Send Modal

namedotget:
- Added safe send modal that allows team safe signers to send funds from the safe
- Refactor


ui/components/safe/SafeExecutionDisclaimer.tsx
```diff
@@ -0,0 +1,149 @@
+import { XMarkIcon } from '@heroicons/react/20/solid'
```
pmoncada, line 1: There's a lot in here - kind of taking it on faith that you've tested sending funds via the team and it's all good. Looks fine here!

ui/components/safe/SafeExecutionDisclaimer.tsx
```diff
@@ -0,0 +1,149 @@
...
+      transaction.dataDecoded?.method === 'Reject Transaction' ||
+      transaction.dataDecoded?.method?.toLowerCase().includes('reject') ||
+      ((transaction.data === '0x' || transaction.data === null) &&
+        ethers.BigNumber.from(transaction.value).eq(0)))
+
+  const method = transaction
+    ? isEthTransfer
+      ? 'Transfer ETH'
+      : isRejectionTx
+      ? 'Reject Transaction'
```
pmoncada, line 48: What other methods are there? Anything else we might want to support? Are we storing nonce values correctly and making sure we don't overwrite?

ui/components/safe/SafeSendModal.tsx
```diff
@@ -0,0 +1,249 @@
...
+    const isAddressValid = to.length === 42 && to.startsWith('0x')
+    const isAmountValid = Number(amount) > 0
+    setIsValid(isAddressValid && isAmountValid)
+  }, [to, amount])
+
+  const handleSend = async (e: React.FormEvent) => {
+    e.preventDefault()
+    if (!to || !amount) {
+      return toast.error('Please fill in all fields')
+    } else if (to.length !== 42 || !to.startsWith('0x')) {
```
pmoncada, line 60: Support ENS?

ui/components/subscription/TeamJobs.tsx
```diff
@@ -80,6 +80,8 @@ export default function TeamJobs({
     }
   }, [router])
 
+  if (!jobs?.[0]) return null
```
pmoncada, line 83: I thought this was done in another PR already?

ui/const/icons.ts
```diff
@@ -0,0 +1,13 @@
+export const COIN_ICONS = {
+  ADA: '/coins/ADA.svg',
```
pmoncada, line 2: Can remove ADA/BNB/BTC (keep WBTC)

pmoncada (approved):

ui/components/safe/SafeExecutionDisclaimer.tsx
namedotget: If the tx has data then I reference the method called on the contract, otherwise if the tx doesnt have data but has an eth value its defined as 'Transfer ETH'. If there is no tx data and no value sent then it gets marked as 'Reject Transaction'.

ui/components/safe/SafeReceiveModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <div className="mt-2 flex flex-col gap-4">
+          <p className="text-lg">
+            {`This Safe can only receive funds on ${DEFAULT_CHAIN_V5.name}.`}
+          </p>
+          <p className="">{safeAddress}</p>
+          <StandardButton
+            className="gradient-2 rounded-full"
+            onClick={() => {
+              navigator.clipboard.writeText(safeAddress)
+              toast.success('Address copied to clipboard')
```
pmoncada, line 54: add a period plz

pmoncada (approved):

PR #495, opened 2025-06-04
Add reserved tokens/payout buttons

mmoncada1:
## Summary
- add new buttons to send reserved tokens and payouts on Mission page
- expose functionality for jbController.sendReservedTokensToSplitsOf and IJBMultiTerminal.sendPayoutsOf

## Testing
- `yarn lint`

------
https://chatgpt.com/codex/tasks/task_e_6840c68ad01083268b9bf0061d8e03b4

### DIFF
--- a/ui/pages/mission/[tokenId].tsx
+++ b/ui/pages/mission/[tokenId].tsx
@@ -231,6 +231,7 @@ export default function MissionProfile({ mission }: ProjectProfileProps) {
       })
 
       const jbTerminalStoreContract = getContract({
+        client: serverClient,
         address: storeAddress,
         abi: IJBTerminalStoreABI.abi as any,
         chain: selectedChain,


ui/pages/mission/[tokenId].tsx
```diff
@@ -505,6 +551,26 @@ export default function MissionProfile({ mission }: ProjectProfileProps) {
                 />
               </div>
             </div>
+            {account && (
+              <div className="px-[5vw] w-full flex justify-center my-4 gap-4">
+                <StandardButton
```
Comment (namedotget) [inline ui/pages/mission/[tokenId].tsx:556]: If the connected wallet is metamask, the chain wont auto-switch so we need to use the PrivyWeb3Button or something similar to check: if the wallet is connected, if it has the current chain (if not it gets added), and if its on the current chain (if not it prompts the user to switch).

PR #535, opened 2025-06-13
footer updates

MitchMcQuinn:
- Modified the expanded footer with a isFullwidth prop to handle both full and fixed-width layouts
- Applied the new expanded footer to the launch page and the homepage


pmoncada:
@namedotget where can I see a preview of the changes on vercel?

namedotget (approved):

PR #536, opened 2025-06-13
fix amount raised

jaderiverstokes:
- switch to JB sdk hook to read eth raised
- wrap ProfileHeader in a component so it has access to JuiceProviders


pmoncada:
good to merge now btw since this is blocking miguel, but would be good to add a test in a future pr

PR #556, opened 2025-06-19
Made wallet pop-out look dope

mmoncada1:


MitchMcQuinn:
Definitely a huge layout improvement. I like the glassmorphism, and I think we should head in that direction, but at the moment it clashes with the rest of the aesthetics. I think we should keep the old background gradient and button color for now, and once we've implemented the changes to the homepage, bring the glassmorphism back

namedotget:
@mmoncada1 looks great, nice work! Can you also update the send modal in PrivyConnectWallet.

mmoncada1:
> @mmoncada1 looks great, nice work! Can you also update the send modal in PrivyConnectWallet.

Just pushed the updated send modal

pmoncada (approved):

pmoncada:
Fix the error with deployment before merging in!

PR #561, opened 2025-06-23
fixed marketplace prices

mmoncada1:


namedotget:
Screenshot 2025-06-23 at 6 06 57 PM
It still looks a little off on 'Ipad Pro'.  Also please make the color of the text and gradient mesh with the rest of the app, you can use something like "gradient-2" check globals.css.

PR #565, opened 2025-06-24
Updates pages with lists for Citizens, teams, and marketplace to use new style

mmoncada1:


namedotget:
Looks good, just make the search bar on the /jobs page consistent with the rest.

PR #590, opened 2025-06-27
Fixed look on $mooney pages to work on full-width

mmoncada1:


pmoncada:
Noticing that the estimate for the MOONEY is super slow, and it's not recognizing my lock amount on other chains / letting me lock despite having MOONEY on there. Aesthetic changes look fine, but we should make sure that the locking / get mooney functionality works even if this change isn't what broke it

namedotget (approved):

pmoncada:
Getting the "estimated MOONEY" when buying is still very slow on my end - I feel like the estimates you get from Uniswap are usually much faster. Any idea what might be causing this?

PR #602, opened 2025-07-01
Fix and optimize retroactive rewards

namedotget:
- Fixed issue w/ ProjectCard not detecting whether the current wallet is a contributor
- Optimized server-side function


namedotget:
> lgtm - there's a lot on our plate rn, but if it's not too insane, it would be good to add a test to check the isContributor v.s. not contributor is working as expected in the future.
> 
> f it's a huge pain to set up that type of test for this then don't worry about it too much though. Just want to make sure this is robust in the future

I will handle this in a separate PR just so we can push the fix. Will need to create tests for RetroactiveRewards and the ProjectCard.

PR #603, opened 2025-07-01
Fixed lock extension across chains

mmoncada1:


pmoncada (commented):
Did you run the following tests manually:

- creating a fresh lock
- pulling out tokens w/ expired lock
- increasing lock time
- increasing lock amount
- doing all the above on each chain (arb, polygon, mainnet, base)

Plz confirm that all works and I'll approve

pmoncada:
Let's just make sure we manually check each of these. I realized we only allow for arbitrum bridging at the moment and it might not be trivial to support bridging for base and polygon

PR #612, opened 2025-07-03
🎨 Launchpad Landing Page Redesign

pmoncada:
Peep the preview. Lots of things changed.


namedotget:
"Scroll to Explore" has a hover effect and makes it feel like its a button, when clicked it would be cool if it scrolled to the next section.  Everything else looks good, just need to fix the merge conflicts.

PR #655, opened 2025-07-15
Fetch wallet balances from etherscan, adjust styling

namedotget:
- Display wallet balances from etherscan, filtering out tokens with an url in the name
- Allow user to send any token in their wallet

### DIFF
--- a/ui/pages/api/etherscan/wallet-tokens.ts
+++ b/ui/pages/api/etherscan/wallet-tokens.ts
@@ -1,0 +1,229 @@
+import { rateLimit } from 'middleware/rateLimit'
+import withMiddleware from 'middleware/withMiddleware'
+import { NextApiRequest, NextApiResponse } from 'next'
+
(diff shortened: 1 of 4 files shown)


ui/pages/api/etherscan/wallet-tokens.ts
```diff
@@ -0,0 +1,194 @@
...
+    const apiKey = process.env.NEXT_PUBLIC_ETHERSCAN_API_KEY
+
+    if (!apiKey) {
+      return res.status(500).json({
+        error: 'Etherscan API key not configured',
+      })
+    }
+
+    // Step 1: Get token transfer events to discover tokens using V2 API
+    const tokentxUrl = `https://api.etherscan.io/v2/api?chainid=${chainId}&module=account&action=tokentx&address=${address}&page=1&offset=10000&sort=desc&apikey=${apiKey}`
```
jaderiverstokes, line 105: can't believe get tokens held by an address is a PRO endpoint. smh.

definitely did not used to be like that

PR #666, opened 2025-07-16
hard code engibob's distribution

jaderiverstokes:


ui/components/nance/RetroactiveRewards.tsx
```diff
@@ -207,10 +220,30 @@ export function RetroactiveRewards({
...
-  const { walletVPs: _vps } = useTotalVPs(addresses)
-  const addressToQuadraticVotingPower = Object.fromEntries(
-    addresses.map((address, index) => [address, _vps[index]])
-  )
+  //const walletVPs = useTotalVPs(addresses)
+  //console.log('walletVPs', walletVPs)
+  //const addressToQuadraticVotingPower = Object.fromEntries(
+  //addresses.map((address, index) => [address, walletVPs[index]])
+  //)
+  const engibobVP = useTotalVP('0x4CBf10c36b481d6afF063070E35b4F42E7Aad201')
```
pmoncada, line 228: Do we want to hardcode this in retroactive rewards .tsx though? Not sure we should merge this into the codebase since it's just a one time patch

pmoncada (changes requested):

PR #685, opened 2025-07-21
Fee UI Review

jaderiverstokes:
- a few copy changes that came up in review but didn't make it into branch merge
- remove progress bar and just show end of week
- bug fix with estimated rewards
- deploy on mainnet


ui/pages/fees.tsx
```diff
@@ -207,11 +207,13 @@ export default function Fees() {
               chain,
             })
 
+            console.log('vMooneyAddress', vMooneyAddress)
```
pmoncada, line 210: Could probably get rid of console.log / debug statements

pmoncada (approved):

PR #690, opened 2025-07-22
added projects section

mmoncada1:


pmoncada (changes requested):
Fixes we discussed IRL

PR #732, opened 2025-08-02
Privy Role Manager (Replacement for GuildXYZ)

namedotget:
- Allow users to verify discord roles directly from the moondao website


pmoncada (approved):
Approved for testing in production. We'll likely want to change where we surface this in the future (i.e. not on the far end of the profile page actions, that might be hard to find), eventually we'll want this to be intuitive and easy to find in the homepage.

PR #757, opened 2025-08-12
XP contracts and api routes

namedotget:
- Adds XP contracts to award XP based on on-chain/off-chain conditions which can be redeemed for an ERC20(MOONEY).
- Endpoints sign EIP-712 proofs and relay XPManager.claimXPFor on behalf of users.

### DIFF
--- a/xp/test/XPFlow.t.sol
+++ b/xp/test/XPFlow.t.sol
@@ -22,7 +22,6 @@ contract XPFlowTest is Test {
     address public user = address(0x123);
     address public owner = address(0x456);
     address public treasury = address(0x789);
-    address public mockTableAddress = address(0xABC);
     address public whitelist = address(0xDEF);
     address public discountList = address(0x111);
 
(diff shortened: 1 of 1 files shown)


xp/.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
+name: CI
```
jaderiverstokes, line 1: i believe this has to go in the root directory. eg moondao/.github/workflows for GitHub to find it.

also you could merge just this file into main so that XP ci will run on this PR

xp/src/XPManager.sol
```diff
@@ -0,0 +1,503 @@
...
+        emit XPEarned(user, amount, userXP[user]);
+    }
+
+    /**
+     * @dev Internal function to update highest threshold reached
+     * @param user Address of the user
+     * @param oldXP Previous XP amount
+     * @param newXP New XP amount
+     */
+    function _updateHighestThresholdReached(address user, uint256 oldXP, uint256 newXP) internal {
```
jaderiverstokes, line 467: ?

delete?

xp/src/verifiers/HasJoinedATeam.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: MIT
```
jaderiverstokes, line 1: curious why a few of these need an oracle. we could check this on chain right?

PR #799, opened 2025-09-05
Fix onboarding funding network issue

namedotget:


ui/components/onboarding/CreateTeam.tsx
```diff
@@ -412,6 +413,7 @@ export default function CreateTeam({ selectedChain, setSelectedTier }: any) {
 
                           return await fundWallet(address, {
                             amount: String(roundedCost),
+                            chain: viemChains[chainSlug],
```
pmoncada, line 416: Will this always default to arb? Or if they're on another chain will this go to that chain? I assume we always want them to be fund on arb?

pmoncada (approved):

PR #807, opened 2025-09-11
Lock pages network selector edits

mmoncada1:


namedotget:
This PR still has the blur effect on real jobs for non-citizens.

PR #813, opened 2025-09-15
Fix Citizen Page not Recognizing Citizen Status

mmoncada1:


ui/pages/citizen/[tokenIdOrName].tsx
```diff
@@ -124,6 +125,9 @@ export default function CitizenDetailPage({ nft, tokenId, hats }: any) {
 
   const isOwner = address?.toLowerCase() === nft?.owner?.toLowerCase()
 
+  // Check if current user is a citizen (for viewing permissions)
+  const currentUserCitizen = useCitizen(selectedChain, citizenContract, address)
```
Comment (namedotget) [inline ui/pages/citizen/[tokenIdOrName].tsx:129]: Use `const {citizen} = useContext(CitizenContext)` to get the citizen nft rather than the hook.

PR #817, opened 2025-09-17
FIxed Data for Proposals on Dashboard

mmoncada1:


mmoncada1:
Screenshot 2025-10-01 at 5 55 25 PM


Maybe we cut it to 2 proposals so that the empty space doesnt appear on the left and right columns, or we can make each card less tall.

PR #834, opened 2025-09-24
Team Creator V3 & Team Member Passthrough Module fix 

namedotget:
- Team Creator V3: Allow authorized signers to create teams for other wallets (gasless/cross-chain), refactor params, add setters
- Fix team member hat passthrough module for teams deployed with team creator v2 so that managers can remove members

### DIFF
--- a/subscription-contracts/script/TeamCreatorSepolia.s.sol
+++ b/subscription-contracts/script/TeamCreatorSepolia.s.sol
@@ -18,17 +18,16 @@ import {MoonDAOTeamCreator} from "../src/MoonDAOTeamCreator.sol";
 import {IHats} from "@hats/Interfaces/IHats.sol";
 
 
-contract TeamCreatorScript is Script {
+contract TeamCreatorSepoliaScript is Script {
     function run() external {
         uint256 deployerPrivateKey = vm.envUint("PRIVATE_KEY");
         address authorizedSignerAddress = vm.envAddress("AUTHORIZED_SIGNER_ADDRESS");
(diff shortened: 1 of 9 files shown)


subscription-contracts/script/TeamCreatorSepolia.s.sol
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: MIT
```
jaderiverstokes, line 1: why do we need a separate sepolia script? can you add a comment

subscription-contracts/src/MoonDAOTeamCreator.sol
```diff
@@ -64,54 +119,74 @@ contract MoonDAOTeamCreator is Ownable {
...
+        HatURIs memory hatURIs,
+        TeamMetadata calldata metadata,
+        address[] memory members
+    ) external payable onlyAuthorizedSigner returns (uint256 tokenId, uint256 childHatId) {
+        return _createMoonDAOTeam(creator, hatURIs, metadata, members);
+    }
 
-        require(whitelist.isWhitelisted(msg.sender) || openAccess, "Only whitelisted addresses can create a MoonDAO Team");
-        
+    function createMoonDAOTeam(
```
jaderiverstokes, line 131: bless you for simplifying the call parameters. this constructor was destroying my compiler 🙏

jaderiverstokes (approved):

subscription-contracts/script/TeamCreatorSepolia.s.sol
namedotget: The config is different for sepolia so having a separate script saves some time.

PR #855, opened 2025-10-03
Add offchain pruning

namedotget:


jaderiverstokes:
not sure if you will need it, but there is PROJECT_HAT_TREE_IDS also.

might not be relevant for the deletion request but just in case.

jaderiverstokes (approved):

PR #881, opened 2025-10-15
Add multichain support to CB Onramp redirect, calc coinbase fees, account for available ETH for mission onramp

namedotget:
- can now add ?chain=base or ?chain=ethereum to mission urls to set the chain (defaults to arbitrum), the url param updates when users switch chains
- arbitrum one is now labeled as (recommended) in the network selector dropdown


jaderiverstokes (approved):

jaderiverstokes:
fyi 429 on the build, maybe rerun it just in case?

PR #948, opened 2025-11-04
Dashboard Mooney Balance Component

namedotget:
- Added mooney balance breakdown component to dashboard header
- Redesigned right side of dashboard header
- Improved tooltip styling and interactions


ryand2d (commented):
Snapshot voting power isn't loading for me on preview.

Not sure if you think this is clearer, but saying Unlocked vs Available, and instead of displaying Locked balance, display vMooney balance (on mouseover we could show the total amount locked + expiration date)?

jaderiverstokes (approved):

PR #968, opened 2025-11-11
Proposal metadata and proposal submission notifications

namedotget:
- Added a discord notification for proposal submission
- Updated metadata for proposals to use title and abstract
- Fixed proposal submission cta linking


ui/pages/api/proposal/new-proposal-notifcation.ts
jaderiverstokes: fyi notifcation -> notification

jaderiverstokes (approved):

PR #971, opened 2025-11-11
Give people a free mint if they contribute to missions, even if they miss the toast

jaderiverstokes:


ui/pages/api/mission/freeMint.ts
```diff
@@ -21,6 +21,14 @@ import { createHSMWallet } from '@/lib/google/hsm-signer'
 import { getChainSlug } from '@/lib/thirdweb/chain'
 import { serverClient } from '@/lib/thirdweb/client'
 
+type CachedBackers = {
+  data: Backer[]
+  timestamp: number
+}
+
+const backersCache: Map<string, CachedBackers> = new Map()
```
namedotget: please use edge caching instead of storing in memory, I added a setCDNCacheHeaders helper, you can check the implementation in /api/tableland/query

namedotget (approved):

PR #1025, opened 2025-12-01
don't default to mainnet when buying mooney

jaderiverstokes:


pmoncada (approved):
Approved! Also a small thing I noticed is that we don't say what the input token is on the page necessarily, you type in e.g. 0.1 but it doesn't say it's "ETH" in an obvious way on the page. Just a small thing that could be added in

PR #1054, opened 2025-12-08
Coinbase onramp create citizen flow

namedotget:
Replace privy onramp with custom coinbase onramp w/ auto tx


namedotget:
> Wasn't able to get past Profile stage of citizen creation

Works for jade and myself, you have to use https://onramp.preview.moondao.com.  The preview listed above wont let you connect a wallet since everything is tied to the onramp.preview domain for that environment, not sure how you got stuck on the "Profile" stage.

PR #1061, opened 2025-12-10
Quests Rework

mmoncada1:
- Introduced on-boarding Quests
- New users must complete each on boarding quest to unlock the rest of the quests
- New Users will see their "current" on boarding quest on their homepage below quick actions
- Users who have completed on boarding will see one featured quest on their homepage
- /quests page reflects these changes as well


ui/lib/xp/config.ts
namedotget: Todo: enforce onboarding quest order in the backend

namedotget:
Screenshot 2025-12-10 at 1 32 33 PM

After claiming the first citizen profile quest the whole component reloaded and then it didn't proceed to the second quest

PR #1064, opened 2025-12-15
EB rewards

namedotget:
# EB Rewards

## Summary
Added EB reward component for managers to the EB team page.  Will eventually open this up publicly after we add rate limiting.

- Rewards are calculated based on 2% of quarterly treasury growth from coinstats
- Component includes collapsible explanation of calculation flow
- API protected by manager authentication middleware
- Supports test environment bypass for Cypress tests


jaderiverstokes (approved):
left a few small comments

PR #1067, opened 2025-12-16
adds contracts/tests for governance

jaderiverstokes:
- new tables
  - Proposals (quarterly votes on proposals + temp check)
  - NonProjectProposals
  - Senators (on chain senator role)
  
  Temp check happens fully on chain, proposal votes are done through tableland and tallied off-chain.


subscription-contracts/src/tables/NonProjectProposal.sol
namedotget: Do you think we'll ever need to add a column? Could make this table dynamic if so

PR #1251, opened 2026-04-09
fix flaky cypress tests

pmoncada:
- Increase cy.wrap timeout for vMOONEY RPC calls (4s -> 30s)
- Handle uncaught Typeform exception in CitizenMetadataModal test when form ID env var is missing in test environment

Made-with: Cursor


jaderiverstokes:
doesn't seem like it actually fixed the tests 😭

mmoncada1 (approved):

PR #1277, opened 2026-04-28
fix bugs

mmoncada1:


jaderiverstokes (commented):
looks good. please add unit tests and test onchain to make sure the layer zero integration works.

PR #1289, opened 2026-05-04
feat: GitHub Actions townhall summarizer (no cookies needed)

mmoncada1:
Adds a scheduled GitHub Actions workflow that runs every Friday at 12:00 UTC to auto-generate town hall summaries. Uses yt-dlp subtitle download from GitHub's runners (not bot-flagged by YouTube), so no cookies are ever needed.


mmoncada1:
Superseded by a full PR on fix-townhall-sum-again which includes this workflow plus all the other fixes.

PR #1316, opened 2026-05-27
EB-002: Add /api/eb/tracker baseline metrics endpoint

mmoncada1:
## Summary
Adds `ui/pages/api/eb/tracker.ts` — a new API endpoint that freezes the EB baseline subscriber counts and exposes live revenue metrics for ongoing tracking.

## What it does
- **Frozen baseline:** Citizens = 199, Teams = 20 (captured 2026-05-27)
- **Subscription costs:** reads `pricePerSecond` from Citizen/Team NFT contracts on Arbitrum (applies `TEAM_DISCOUNT = 0.067` to match `arr.ts`)
- **Trailing-30d revenue:** canonical on-chain pipeline (same as `/api/eb/audit`)
- **Trailing-365d revenue:** preferred run-rate metric for annual subscriptions

## Verification
```bash
curl https://www.moondao.com/api/eb/tracker | jq
```

Ref: EB-002

### DIFF
--- a/ui/pages/api/eb/tracker.ts
+++ b/ui/pages/api/eb/tracker.ts
@@ -1,0 +1,170 @@
+/**
+ * EB Tracker — Baseline Metrics Snapshot
+ *
+ * Freezes the baseline subscriber counts captured on 2026-05-27
+ * (Citizens 199, Teams 20) alongside live trailing-30-day revenue
+ * and per-tier monthly subscription costs, so the EB can track
+ * growth against a fixed reference point.
+ *
+ * Usage:  GET /api/eb/tracker
+ *
+ * All revenue figures come from the same canonical on-chain pipeline
+ * used by /api/eb/audit — Etherscan v2 internal txs on Arbitrum.
+ */
+import { NextApiRequest, NextApiResponse } from 'next'
+import { getCanonicalSubscriptionRevenue } from '@/lib/treasury/canonicalRevenue'
+import { getETHPrice } from '@/lib/etherscan'
+import CitizenABI from 'const/abis/Citizen.json'
+import TeamABI from 'const/abis/Team.json'
+import { CITIZEN_ADDRESSES, TEAM_ADDRESSES } from 'const/config'
+import { getContract, readContract } from 'thirdweb'
+import { serverClient } from '@/lib/thirdweb/client'
+import { arbitrum } from '@/lib/rpc/chains'
+import { getChainSlug } from '@/lib/thirdweb/chain'
+
+// ─── Frozen baseline (captured 2026-05-27) ───────────────────────────────────
+const BASELINE = {
+  capturedAt: '2026-05-27',
+  citizens: 199,
+  teams: 20,
+} as const
+
+const SECONDS_PER_MONTH = 30 * 24 * 60 * 60 // 30-day month approximation
+
+// arr.ts applies this factor to the team contract's raw pricePerSecond because
+// the team subscription UX charges ~0.0333 ETH/yr (not the ~0.493 ETH/yr the
+// raw pricePerSecond would imply at face value).  Without it the displayed cost
+// is 14.8× too high.  Keep in sync with TEAM_DISCOUNT in lib/treasury/arr.ts.
+const TEAM_PRICE_DISCOUNT = 0.067
+
+async function getSubscriptionPricesETH(): Promise<{
+  citizenPerMonth: number
+  teamPerMonth: number
+}> {
+  try {
+    const chain = arbitrum
+    const chainSlug = getChainSlug(chain)
+
+    const citizenContract = getContract({
+      client: serverClient,
+      address: CITIZEN_ADDRESSES[chainSlug],
+      chain,
+      abi: CitizenABI as any,
+    })
+    const teamContract = getContract({
+      client: serverClient,
+      address: TEAM_ADDRESSES[chainSlug],
+      chain,
+      abi: TeamABI as any,
+    })
+
+    const [citizenPricePerSecond, teamPricePerSecond] = await Promise.all([
+      readContract({ contract: citizenContract, method: 'pricePerSecond' }),
+      readContract({ contract: teamContract, method: 'pricePerSecond' }),
+    ])
+
+    return {
+      citizenPerMonth: (Number(citizenPricePerSecond) * SECONDS_PER_MONTH) / 1e18,
+      teamPerMonth: (Number(teamPricePerSecond) * SECONDS_PER_MONTH * TEAM_PRICE_DISCOUNT) / 1e18,
+    }
+  } catch {
+    // Fallback to known values if RPC unavailable
+    // citizen: 0.0111 ETH/year → 0.0111×(30/365) = 0.000912 ETH/month
+    // team:    0.0333 ETH/year → 0.0333×(30/365) = 0.002737 ETH/month
+    return {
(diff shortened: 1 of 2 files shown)


ui/pages/api/eb/tracker.ts
```diff
@@ -0,0 +1,170 @@
...
+    const [ethPrice, prices] = await Promise.all([
+      getETHPrice(),
+      getSubscriptionPricesETH(),
+    ])
+
+    // Both windows share the same memoised Etherscan fetch — no extra API calls.
+    const [trailing30d, trailing365d] = await Promise.all([
+      getCanonicalSubscriptionRevenue(trailing30dStart,  now, ethPrice),
+      getCanonicalSubscriptionRevenue(trailing365dStart, now, ethPrice),
+    ])
```
pmoncada: The comment says "Both windows share the same memoised Etherscan fetch — no extra API calls," but that only holds on a *warm* cache. `fetchAllInternalTxs` in `canonicalRevenue.ts` has no in-flight dedup — it only checks `memo` at the top. Because these two calls run concurrently in `Promise.all`, on a cold cache both enter `fetchAllInternalTxs`, both see `memo === null`, and both issue the identical Etherscan request (the second's result just overwrites `memo`). So the first request to this endpoint after a cold start / >10min idle makes 2 concurrent Etherscan calls, which on the free 5 req/s tier can actually trip the `NOTOK` rate-limit retry path. Easiest fixes: `await` the two windows sequentially (the second then reuses `memo`), or add an in-flight promise guard in `fetchAllInternalTxs` (like `useOnrampRegion` does).

ui/pages/api/eb/tracker.ts
```diff
@@ -0,0 +1,170 @@
...
+  teams: 20,
+} as const
+
+const SECONDS_PER_MONTH = 30 * 24 * 60 * 60 // 30-day month approximation
+
+// arr.ts applies this factor to the team contract's raw pricePerSecond because
+// the team subscription UX charges ~0.0333 ETH/yr (not the ~0.493 ETH/yr the
+// raw pricePerSecond would imply at face value).  Without it the displayed cost
+// is 14.8× too high.  Keep in sync with TEAM_DISCOUNT in lib/treasury/arr.ts.
+const TEAM_PRICE_DISCOUNT = 0.067
```
pmoncada: `TEAM_PRICE_DISCOUNT = 0.067` is a copy of `TEAM_DISCOUNT` in `lib/treasury/arr.ts`, and the comment correctly flags they must stay in sync. Rather than relying on a comment, consider `export`ing the constant from `arr.ts` and importing it here so they can't silently drift. This is exactly the kind of magic factor that gets updated in one place and forgotten in the other.

ui/pages/api/eb/tracker.ts
```diff
@@ -0,0 +1,170 @@
...
+    const [citizenPricePerSecond, teamPricePerSecond] = await Promise.all([
+      readContract({ contract: citizenContract, method: 'pricePerSecond' }),
+      readContract({ contract: teamContract, method: 'pricePerSecond' }),
+    ])
+
+    return {
+      citizenPerMonth: (Number(citizenPricePerSecond) * SECONDS_PER_MONTH) / 1e18,
+      teamPerMonth: (Number(teamPricePerSecond) * SECONDS_PER_MONTH * TEAM_PRICE_DISCOUNT) / 1e18,
+    }
+  } catch {
```
pmoncada: This `catch` is silent (no `console.error`), unlike the equivalent fallback in `arr.ts`'s `getSANContractPrices`. Worse, the fallback returns hardcoded prices (`0.000912` / `0.002737`) that look like real on-chain data in the response — there's no flag in the JSON telling a consumer the prices are stale/estimated. For an endpoint whose whole purpose is auditable EB metrics, I'd at least log the error and surface something like `pricesSource: 'fallback'` so a reader can tell when the RPC read failed.

ui/pages/api/eb/tracker.ts
```diff
@@ -0,0 +1,170 @@
...
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  try {
+    const now = Date.now()
+    const trailing30dStart  = now - 30  * 24 * 60 * 60 * 1000
+    const trailing365dStart = now - 365 * 24 * 60 * 60 * 1000
+
+    const [ethPrice, prices] = await Promise.all([
+      getETHPrice(),
+      getSubscriptionPricesETH(),
+    ])
```
pmoncada, line 101: If `getETHPrice()` fails it returns `0` (see `lib/etherscan` / the `arr.ts` precedent), in which case every `*USD` field in the response silently becomes `0` while the ETH figures look fine. That's a confusing failure mode for a metrics endpoint. Consider detecting `ethPrice === 0` and either returning a 502/`error` or omitting the USD fields rather than emitting `$0`.

ui/pages/api/eb/tracker.ts
```diff
@@ -0,0 +1,170 @@
...
+import { getChainSlug } from '@/lib/thirdweb/chain'
+
+// ─── Frozen baseline (captured 2026-05-27) ───────────────────────────────────
+const BASELINE = {
+  capturedAt: '2026-05-27',
+  citizens: 199,
+  teams: 20,
+} as const
+
+const SECONDS_PER_MONTH = 30 * 24 * 60 * 60 // 30-day month approximation
```
pmoncada: Minor/heads-up: `SECONDS_PER_MONTH` uses a 30-day month, so `baselineMonthlyTotal × 12` = 360 days, which is ~1.4% short of a true annual figure (and `arr.ts` annualizes with 365.25 days). Fine as long as nobody downstream multiplies this monthly total by 12 to get an annual run-rate; worth a one-line note if that's a likely use.

ui/pages/api/eb/tracker.ts
```diff
@@ -0,0 +1,170 @@
...
+      getCanonicalSubscriptionRevenue(trailing30dStart,  now, ethPrice),
+      getCanonicalSubscriptionRevenue(trailing365dStart, now, ethPrice),
+    ])
+
+    // Monthly cost for ALL baseline subscribers at current on-chain prices
+    const monthlySubscriptionCostETH =
+      BASELINE.citizens * prices.citizenPerMonth +
+      BASELINE.teams * prices.teamPerMonth
+
+    res.status(200).json({
```
pmoncada, line 136: Minor: this is an unauthenticated public endpoint that fans out to Etherscan + two RPC reads on every cold-cache hit. Since the underlying data only changes slowly and is already memoised for 10min, consider setting a `Cache-Control` / `s-maxage` header (e.g. `res.setHeader('Cache-Control', 's-maxage=600, stale-while-revalidate')`) so repeated EB/dashboard polls are served from the CDN instead of re-invoking the function.

pmoncada (commented):
Clean, self-contained endpoint — reusing the canonical on-chain revenue pipeline (the same one `/api/eb/audit` and `/api/eb/rewards` use) keeps the EB numbers consistent and reproducible, and matching `arr.ts`'s team discount + citizen/team price handling is the right call. The frozen-baseline approach is well documented. A few mostly-minor things below, plus one comment that's slightly inaccurate about API-call dedup. Nothing blocking.

pmoncada:
did not build, pls fix

mmoncada1:
Closing — recreating this PR from the branch on Official-MoonDao instead of a fork. Superseded by the new PR.

PR #1317, opened 2026-05-28
feat: Coinbase Headless Onramp (Apple Pay iframe, US only)

mmoncada1:
- Add CBHeadlessOnramp component with stepped phone verify + pay flow
- Add CBOnrampModal dual-path router (headless for US, legacy for non-US)
- Add usePhoneVerification hook (Privy SMS OTP, 60-day staleness check)
- Add useOnrampRegion hook (geo detection, safe fallback to legacy)
- Add /api/coinbase/create-order endpoint (CDP v2 headless orders)
- Add /api/coinbase/region endpoint (Vercel geo header detection)

Fixes: iframe sandbox allow-popups-to-escape-sandbox for Apple Pay sheet
Fixes: remove referrerPolicy=no-referrer (breaks Coinbase domain validation)
Fixes: treat Privy accounts with no verifiedAt as fresh (not stale)

### DIFF
--- a/ui/components/coinbase/CBHeadlessOnramp.tsx
+++ b/ui/components/coinbase/CBHeadlessOnramp.tsx
@@ -18,6 +18,13 @@ import {
   LargeAmountExchangeOnrampNotice,
 } from './LargeAmountExchangeOnrampNotice'
 import usePhoneVerification from '@/lib/coinbase/usePhoneVerification'
+import {
+  type HeadlessStatus,
+  getQuoteNetworkName,
+  getOnrampNetworkName,
+  parseOnrampMessage,
+  mapOnrampEvent,
+} from '@/lib/coinbase/headlessEvents'
 
 interface CBHeadlessOnrampProps {
   address: string
@@ -43,84 +50,6 @@ interface CBHeadlessOnrampProps {
 
 const MOCK_ONRAMP = process.env.NEXT_PUBLIC_MOCK_ONRAMP === 'true'
 
-// Map post-message event names to internal UI states
-type HeadlessStatus =
-  | 'idle' // before user clicks Buy
-  | 'creating' // POST /create-order in flight
-  | 'iframe-loading' // iframe mounted, waiting for load_success
-  | 'ready' // pay button visible inside the iframe
-  | 'committing' // user pressed Apple Pay, awaiting commit_success
-  | 'polling' // commit succeeded, waiting for settlement
-  | 'success'
-  | 'error'
-
-interface PostMessageData {
-  eventName?: string
-  data?: { errorCode?: string; errorMessage?: string }
-}
-
-function getQuoteNetworkName(chain: any): string {
-  const chainName = chain?.name?.toLowerCase() || 'ethereum'
-  const chainId = chain?.id
-  switch (chainName) {
-    case 'base':
-    case 'base sepolia':
-    case 'base sepolia testnet':
-      return 'base'
-    case 'polygon':
-      return 'polygon'
-    default:
-      switch (chainId) {
-        case 8453:
-        case 84532:
-          return 'base'
-        case 137:
-          return 'polygon'
-        default:
-          return 'ethereum'
-      }
-  }
-}
-
-function getOnrampNetworkName(chain: any): string {
-  const chainName = chain?.name?.toLowerCase() || 'ethereum'
-  const chainId = chain?.id
-  switch (chainName) {
-    case 'arbitrum':
-    case 'arbitrum one':
-    case 'arbitrum sepolia':
-      return 'arbitrum'
-    case 'base':
-    case 'base sepolia':
-    case 'base sepolia testnet':
-      return 'base'
-    case 'polygon':
-      return 'polygon'
-    case 'optimism':
-      return 'optimism'
-    case 'ethereum':
-    case 'mainnet':
-    case 'sepolia':
-      return 'arbitrum'
-    default:
-      switch (chainId) {
-        case 42161:
-        case 421614:
-          return 'arbitrum'
-        case 8453:
-        case 84532:
-          return 'base'
-        case 137:
-          return 'polygon'
-        case 10:
-        case 11155420:
-          return 'optimism'
-        default:
-          return 'ethereum'
-      }
-  }
-}
-
 export const CBHeadlessOnramp: React.FC<CBHeadlessOnrampProps> = ({
   address,
   selectedChain,
@@ -142,6 +71,7 @@ export const CBHeadlessOnramp: React.FC<CBHeadlessOnrampProps> = ({
     phoneNumber,
     isLinked: hasPhone,
     isStale: phoneStale,
+    verifiedAt: phoneVerifiedAt,
     requestVerification,
     refreshVerification,
   } = usePhoneVerification()
@@ -167,6 +97,13 @@ export const CBHeadlessOnramp: React.FC<CBHeadlessOnrampProps> = ({
   const [debouncedEthAmount, setDebouncedEthAmount] = useState(ethAmount)
 
   const iframeRef = useRef<HTMLIFrameElement | null>(null)
+
+  // Track latest status in a ref so the postMessage listener can read it
+  // without re-subscribing on every status change.
+  const statusRef = useRef<HeadlessStatus>(status)
+  useEffect(() => {
+    statusRef.current = status
+  }, [status])
 
   useEffect(() => {
     if (!allowAmountInput) {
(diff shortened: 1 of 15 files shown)


ui/components/coinbase/CBHeadlessOnramp.tsx
```diff
@@ -0,0 +1,840 @@
...
+          )}
+
+          <iframe
+            ref={iframeRef}
+            data-testid="cbheadless-iframe"
+            src={paymentLinkUrl}
+            title="Coinbase Onramp"
+            allow="payment"
+            sandbox="allow-scripts allow-same-origin"
+            referrerPolicy="no-referrer"
```
pmoncada, line 509: These iframe attributes seem to contradict two of the fixes called out in the PR description:

- The description says *"Fixes: iframe sandbox allow-popups-to-escape-sandbox for Apple Pay sheet"*, but the committed `sandbox` is just `allow-scripts allow-same-origin` — no `allow-popups` / `allow-popups-to-escape-sandbox` (and no `allow-forms`). The Apple Pay / 3DS sheet typically needs to open a popup, so this looks like it would block it.
- The description says *"Fixes: remove referrerPolicy=no-referrer (breaks Coinbase domain validation)"*, but `referrerPolicy="no-referrer"` is still here. If Coinbase validates the embedding domain via the `Referer` header, stripping it will break that validation.

Can you double-check the committed state matches the intended fixes? As written, the code looks like the *pre-fix* version of both.

pmoncada (commented):
Really nice piece of work — this is a well-structured response to Coinbase deprecating guest-checkout (debit card / Apple Pay) on the hosted onramp as of June 30, 2026. Splitting the pure logic (`headlessEvents.ts`, `headlessOrder.ts`) out for unit testing, the region-gated dual-path router with a safe fallback to the legacy logged-in flow, and the 60-day phone re-verification handling are all solid. The `sanitizeClientIp` helper in particular is thorough.

My main concerns are around (1) trusting payment success purely from a client-side `postMessage` with no origin check, (2) the committed iframe attributes appearing to contradict two of the fixes listed in the PR description (Apple Pay sheet + Coinbase domain validation), and (3) the server-side US-only gate defaulting to *allow* when geo headers are missing. Details inline. Since this touches real money movement, I'd want the success-verification and origin-validation points resolved before merge.

ui/components/coinbase/CBHeadlessOnramp.tsx
```diff
@@ -0,0 +1,840 @@
...
+  ])
+
+  // ---------------------------------------------------------------------------
+  // Post-message listener for the embedded payment-link iframe.
+  // Docs: onramp_api.{load_pending,load_success,load_error,commit_success,
+  //                  commit_error,cancel,polling_start,polling_success,polling_error}
+  // ---------------------------------------------------------------------------
+  useEffect(() => {
+    if (!paymentLinkUrl) return
+    const handler = (event: MessageEvent) => {
```
pmoncada, line 242: This `message` handler doesn't validate `event.origin` or `event.source`. Any script in any window (another tab embedding our page, an extension, the iframe's own subframes, etc.) can `postMessage` a `{ eventName: 'onramp_api.polling_success' }` payload and we'll treat it as a real payment success. The comment notes the origin can't be pinned because sandbox uses a different host, but you can still (a) maintain an allowlist of known Coinbase hosts (`pay.coinbase.com`, the sandbox host, etc.) and (b) require `event.source === iframeRef.current?.contentWindow`. Without that, the success path is spoofable.

ui/components/coinbase/CBHeadlessOnramp.tsx
```diff
@@ -0,0 +1,840 @@
...
+      // We can't always pin to a Coinbase origin (sandbox uses different host),
+      // so we parse defensively and only act on well-formed onramp_api.* messages.
+      const parsed = parseOnrampMessage(event.data)
+      const result = mapOnrampEvent(parsed, statusRef.current)
+      if (result.ignored) return
+
+      if (result.status) setStatus(result.status)
+      if (result.error !== undefined) setError(result.error)
+      if (result.errorCode !== undefined) setErrorCode(result.errorCode)
+      if (result.fireSuccess) {
```
pmoncada, line 252: Payment success is trusted entirely from a client-side `postMessage` (`onramp_api.polling_success`) — there's no server-side confirmation that the order actually settled. Combined with the missing origin check above, this means `onPaymentSuccess` (and the downstream `onrampSuccess=true` / auto-transaction in the modal) can fire without a real payment. Since `create-order` returns an `orderId`, consider verifying order status server-side (poll the CDP order) before firing the success callback, and treat the `postMessage` only as a UI hint.

ui/components/coinbase/CBHeadlessOnramp.tsx
```diff
@@ -0,0 +1,840 @@
...
+        const response = await fetch('/api/coinbase/buy-quote', {
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json' },
+          credentials: 'include',
+          body: JSON.stringify({
+            paymentAmount: initialEstimateUSD,
+            destinationAddress: address,
+            purchaseNetwork: quoteNetwork,
+            purchaseCurrency: 'ETH',
+            paymentMethod: 'APPLE_PAY',
```
pmoncada, line 174: The preview quote is fetched with `paymentMethod: 'APPLE_PAY'`, but the actual order is created with `GUEST_CHECKOUT_APPLE_PAY` (different fee schedule / limits on Coinbase's side). The total shown to the user here — and used for `quoteData.purchaseAmount`, which is then passed as `paymentAmount` to `create-order` — may not match what guest checkout actually charges. Worth confirming the guest-checkout quote matches, or requesting the quote with the same method the order uses.

ui/components/coinbase/CBHeadlessOnramp.tsx
```diff
@@ -0,0 +1,840 @@
...
+    try {
+      setStatus('creating')
+      setError(null)
+      setErrorCode(null)
+
+      if (onBeforeNavigate) {
+        try {
+          await onBeforeNavigate()
+          // Allow any localStorage writes (e.g. the resume-JWT) to flush.
+          await new Promise((r) => setTimeout(r, 100))
```
pmoncada, line 303: Minor: `localStorage` writes are synchronous, so the `await new Promise((r) => setTimeout(r, 100))` to "let localStorage flush" shouldn't be necessary — it just adds 100ms of latency before every order. If there's a real async dependency here (e.g. the JWT is written from an async callback inside `onBeforeNavigate`), it'd be better to await that directly rather than sleeping.

ui/pages/api/coinbase/create-order.ts
```diff
@@ -0,0 +1,128 @@
...
+    // In sandbox/mock mode prefix the partnerUserRef so Coinbase short-circuits
+    // the payment without charging the card.
+    const effectivePartnerUserRef = resolvePartnerUserRef(
+      partnerUserRef as string,
+      MOCK_ONRAMP
+    )
+
+    // Detect country for region restrictions (Headless = US only). Coinbase
+    // determines the precise region itself from clientIp; we only gate here.
+    const country = getCountryFromHeaders(req) || 'US'
```
pmoncada, line 55: `getCountryFromHeaders(req) || 'US'` defaults to **allowing** the headless flow when the geo header is absent. `getCountryFromHeaders` returns `null` whenever neither the Vercel nor Cloudflare country header is present (local dev, non-Vercel/CF deploys, or a proxy that strips it). Since the Headless Onramp is contractually US-only, the server gate — which is the authoritative one — should default-deny (`|| null`, i.e. reject when unknown) rather than default-allow. The client `useOnrampRegion` already falls back to the legacy flow on uncertainty; the server should be at least as strict.

ui/pages/api/coinbase/create-order.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    // Append sandbox query param for local/mock testing per docs
+    if (paymentLinkUrl) {
+      paymentLinkUrl = applySandboxParam(paymentLinkUrl, paymentMethod, MOCK_ONRAMP)
+    }
+
+    return res.status(200).json({
+      paymentLinkUrl,
+      orderId,
+      raw: data,
```
pmoncada, line 120: Returning `raw: data` (the full CDP order response) to the browser exposes Coinbase's internal order payload to the client. Unless the frontend needs specific fields beyond `paymentLinkUrl`/`orderId`, consider dropping `raw` (or gating it behind `MOCK_ONRAMP`) to avoid leaking internal/PII-adjacent fields.

ui/lib/coinbase/headlessEvents.ts
```diff
@@ -0,0 +1,175 @@
...
+      return 'arbitrum'
+    case 'base':
+    case 'base sepolia':
+    case 'base sepolia testnet':
+      return 'base'
+    case 'polygon':
+      return 'polygon'
+    case 'optimism':
+      return 'optimism'
+    case 'ethereum':
```
pmoncada, line 61: Just confirming intent: `ethereum` / `mainnet` / `sepolia` map to `arbitrum` for the destination network, while `getQuoteNetworkName` prices Arbitrum quotes on `'ethereum'`. So a user who selects "Ethereum" actually receives ETH on Arbitrum. That's presumably intentional (MoonDAO settles on Arbitrum, per the comment), but it's a surprising silent network swap — worth making sure the UI clearly tells the user which network they'll actually receive funds on, since the quote summary shows `selectedChain?.name`.

mmoncada1:
Closing in favour of a clean PR from the same branch — all commits are identical, just removing the noisy fork history from the diff view.

PR #1318, opened 2026-05-28
Add Complete profile notice and claimable quests notice at top of dashboard

mmoncada1:


ui/components/home/SignedInDashboard.tsx
```diff
@@ -125,6 +128,46 @@ function getCitizenMetadata(citizen: any): string | null {
   return null
 }
 
+// Determine which key profile fields a citizen is still missing
+function getMissingProfileFields(citizen: any): string[] {
```
pmoncada, line 136: The completeness definition here doesn't match the on-chain `HAS_COMPLETED_CITIZEN_PROFILE` verifier. Per `ui/lib/xp/config.ts` that quest considers a profile complete with **description + (location OR discord OR twitter OR website)**. This function instead requires photo **AND** bio **AND** location **AND** a social link — all four. As a result a citizen can have already claimed the "Citizen Profile" XP (MoonDAO considers them complete) yet still see "Profile incomplete" here (e.g. they have a bio + twitter but no location/photo). Consider aligning this with the verifier's definition, or relabel the notice so it's clearly "enhance your profile" rather than implying it's incomplete.

ui/components/home/SignedInDashboard.tsx
```diff
@@ -338,10 +381,23 @@ export default function SignedInDashboard({
   // Citizen metadata modal state
   const [citizenMetadataModalEnabled, setCitizenMetadataModalEnabled] = useState(false)
 
+  // Incomplete profile notice state
+  const [profileNoticeDismissed, setProfileNoticeDismissed] = useState(false)
```
pmoncada: `profileNoticeDismissed` is component state only, so the dismissal doesn't persist — the notice reappears on every navigation back to the dashboard / remount. Consider persisting per-citizen in `localStorage` (and/or auto-hiding once the profile is actually completed) so dismiss feels like dismiss.

ui/components/home/SignedInDashboard.tsx
```diff
@@ -338,10 +381,23 @@ export default function SignedInDashboard({
...
+  const missingProfileFields = getMissingProfileFields(citizen)
+  const showProfileNotice =
+    !!citizen && missingProfileFields.length > 0 && !profileNoticeDismissed
+  const profileEditHref =
```
pmoncada, line 391: `profileEditHref` duplicates the exact href logic already inlined in the existing Edit Profile link below (`.../?edit=1` with the `/join` fallback). Now that you have this variable, reuse it there too instead of repeating the `generatePrettyLinkWithId(...)` ternary.

ui/lib/xp/useClaimableQuestsCount.ts
```diff
@@ -0,0 +1,98 @@
...
+
+          try {
+            const res = await fetch(
+              `${v.route}?user=${userAddress}&accessToken=${accessToken}`
+            )
+            if (!res.ok) return false
+            const data = await res.json()
+            if (data.error) return false
+
+            const metric = data[v.metricKey]
```
pmoncada: Re-deriving eligibility from `metricKey` (`metric > 0`) duplicates—and can diverge from—the threshold logic the API already applies. The GET proof routes return an explicit `eligible` boolean (see `has-voted-proof.ts`, which returns `eligible: false` when `votesCount < VOTES_THRESHOLD`). For any verifier whose threshold is > 1, `metric > 0` will over-count claimable quests. Suggest just using `data.eligible` and dropping the per-type metric parsing.

ui/lib/xp/useClaimableQuestsCount.ts
```diff
@@ -0,0 +1,98 @@
...
+            return false
+          }
+          if (claimed) return false
+
+          // 2. Check eligibility via the quest's backend API
+          if (!v.route || !v.metricKey || !accessToken) return false
+
+          try {
+            const res = await fetch(
+              `${v.route}?user=${userAddress}&accessToken=${accessToken}`
```
pmoncada: Passing the Privy `accessToken` as a query-string param means it can end up in server access logs, proxies, and browser history. The routes also accept it via POST body (`getUserAndAccessToken` reads body for non-GET). Prefer POST with the token in the body, or at minimum be aware of the logging exposure here.

ui/lib/xp/useClaimableQuestsCount.ts
```diff
@@ -0,0 +1,98 @@
...
+      } catch {
+        // Not authenticated — skip
+      }
+
+      const checks = await Promise.allSettled(
+        XP_VERIFIERS.map(async (v) => {
+          // 1. Check on-chain: has the user already claimed this verifier?
+          let claimed = false
+          try {
+            claimed = Boolean(await readContract({
```
pmoncada: This fires an on-chain `readContract` for every verifier (currently 8) on each dashboard mount, before the `accessToken` guard on line 59 — so even unauthenticated users incur all the contract reads only to bail out. Consider short-circuiting when there's no `accessToken`, and/or batching via multicall, to avoid a burst of RPC calls on every render of the home page.

pmoncada (commented):
Nice addition — the incomplete-profile nudge and the claimable-quests pill are a good way to drive engagement from the dashboard. A few things worth tightening before merge, mostly around keeping the "profile complete" definition consistent with the on-chain verifier, persisting the dismissal, and reducing redundant network work in the new hook. Details inline.

pmoncada:
check failing test

pmoncada (approved):

PR #1332, opened 2026-05-29
revert(onramp): roll back headless onramp (#1320) + hotfix (#1331), keep legacy

mmoncada1:
## Summary

Reverts the Coinbase Headless Onramp work and its prod hotfix. We're keeping only the legacy onramp for now while we assess options.

Reverts:
- **#1320** — feat: Coinbase Headless Onramp (Apple Pay iframe, US only)
- **#1331** — fix(onramp): don't block US users when geo header is absent (prod hotfix)

## Why

The Headless Onramp's only functional payment method is **Apple Pay**, which on the web works **only in Safari** (macOS/iOS). Non-Safari users hit `ERROR_CODE_GUEST_APPLE_PAY_NOT_SUPPORTED` and dead-end. `GUEST_CHECKOUT_GOOGLE_PAY` exists in Coinbase's enum but is **not yet functional** (API note: "This API currently only supports `GUEST_CHECKOUT_APPLE_PAY`"). There is no direct card-entry form — cards only work if added to Apple Wallet on Safari.

Given that, the headless flow doesn't broaden coverage enough to justify shipping now. Reverting to legacy keeps the onramp working for **all** browsers until we decide on a path forward (wait for Coinbase Google Pay, or add a non-Coinbase fallback before the legacy guest-checkout deprecation on 2026-06-30).

## What this removes

- `CBHeadlessOnramp` component + headless libs (`headlessOrder`, `headlessEvents`, `useOnrampRegion`, `usePhoneVerification`)
- `create-order` / `region` API routes
- Headless routing in `CBOnrampModal` and `MissionContributeModal` (back to legacy `CBOnramp`)
- Temporary `/onramp-test` probe page
- Related unit/integration tests and sandbox script
- CSP additions for the headless flow

## Net effect

Onramp everywhere reverts to the **legacy** Coinbase guest-checkout flow (works in all browsers). No functional onramp is lost for users.


pmoncada (approved):
wait for review to go through but approving rn

PR #1336, opened 2026-06-01
Fix Weekly Check in

mmoncada1:
## Summary

Makes the weekly liquidity-reward **check-in** flow reliable across Ethereum, Arbitrum, and Base. Previously users could see a green "successfully checked in" toast while **nothing actually changed on-chain**, the button stayed stuck on "Check In", and repeated presses re-prompted the wallet with no-op transactions.

Affects both the dashboard widget (`ui/components/tokens/WeeklyRewardPool.tsx`) and the standalone fees page (`ui/pages/fees.tsx`).

## Root causes & fixes

1. **`underlying network changed` (ethers v5)**
   The thirdweb `account` from `useActiveAccount()` is bound to the chain selected at render time. After switching chains mid-loop, ethers' cached network no longer matched the wallet. → Re-derive the account from the wallet's **current** ethers provider after every switch.

2. **`nonce has already been used` shown as a hard error, then as fake success**
   Treating nonce collisions as blind success masked real failures. → Success is now gated **solely** on on-chain verification: `lastCheckIn[address] === weekStart` (with short retries for RPC propagation). A transaction that doesn't land can no longer report success.

3. **Failed automatic chain switch (e.g. to Ethereum)**
   The old 5s wait + unguarded `switchChain` was too fragile for injected wallets (MetaMask requires user approval). → New `switchToChain()` helper: retries, a 20s approval window, settle delay, and user-reject detection.

4. **Stale button / repeated wallet prompts**
   `isCheckedIn` is now derived from `feeData`, and on-chain status is refreshed after check-in, so the button reliably flips to "Checked In ✓" and stops re-prompting. Eligible chains are also optimistically marked to prevent no-op re-sends.

5. **Unhelpful "No check-ins were submitted" message**
   The real per-chain failure reason (revert/RPC message, user-declined switch, or "did not register on-chain") now surfaces in the toast.

## Diagnostic tooling

Adds `ui/scripts/check-checkin-status.mjs` — reads `weekStart`, `lastCheckIn[address]`, and vMOONEY balance directly from public RPCs to confirm real on-chain check-in status per chain. Standalone script, not imported anywhere (no build/bundle impact).

```bash
node ui/scripts/check-checkin-status.mjs 0xYOURADDRESS
```

## Testing

Verified end-to-end against mainnet with a real wallet:
- Arbitrum check-in registered on-chain (`lastCheckIn === weekStart` ✅).
- Ethereum auto-switch + check-in registered on-chain ✅.
- Base correctly skipped (0 vMOONEY balance).
- Button flips to "Checked In ✓" once all eligible chains are verified; no more phantom successes or repeat prompts.

## Notes

- No gas overrides are set anywhere; wallets use standard network estimates.
- `checkIn()` is a storage-only call, so wallets show "No changes" in the asset-diff preview — this is expected and not an error.

### DIFF
--- a/ui/components/tokens/WeeklyRewardPool.tsx
+++ b/ui/components/tokens/WeeklyRewardPool.tsx
@@ -7,7 +7,7 @@ import { FEE_HOOK_ADDRESSES } from 'const/config'
 import { BigNumber } from 'ethers'
 import { ethers } from 'ethers'
 import Link from 'next/link'
-import { useContext, useState, useEffect, useMemo } from 'react'
+import { useContext, useState, useEffect, useMemo, useCallback } from 'react'
 import toast from 'react-hot-toast'
 import { getContract, readContract, prepareContractCall, sendAndConfirmTransaction } from 'thirdweb'
 import { ethers5Adapter } from 'thirdweb/adapters/ethers5'
@@ -184,16 +184,18 @@ export default function WeeklyRewardPool() {
     fetchEstimates()
   }, [address, chains, WEEK])
 
-  useEffect(() => {
+  // Reusable on-chain status loader. Exposed as a callback so it can be
+  // re-run after a check-in to reconcile the UI with the real on-chain state
+  // (the txs are already confirmed at that point, so lastCheckIn is updated).
+  const refreshFeeData = useCallback(async () => {
     if (!address) return
 
-    const fetchStatus = async () => {
-      // Each chain is fetched independently. A single chain failure (e.g.
-      // public RPC hiccup or rate limit) used to reject the whole Promise.all
-      // and leave `feeData` empty, which surfaced as "No fee data available"
-      // when the user tried to check in. Isolate failures per-chain so the
-      // remaining chains still populate.
-      const raw = await Promise.all(
+    // Each chain is fetched independently. A single chain failure (e.g.
+    // public RPC hiccup or rate limit) used to reject the whole Promise.all
+    // and leave `feeData` empty, which surfaced as "No fee data available"
+    // when the user tried to check in. Isolate failures per-chain so the
+    // remaining chains still populate.
+    const raw = await Promise.all(
         chains.map(async (chain) => {
           const slug = getChainSlug(chain)
           try {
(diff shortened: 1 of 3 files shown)


pmoncada:
## Review

Solid, well-reasoned fix. The diagnosis (stale ethers v5 signer network cache after `switchChain`, plus trusting the tx promise instead of on-chain state) is correct, and the fixes address root causes rather than symptoms. The `getEthersProvider()` + `ethers5Adapter.signer.fromEthers()` pattern matches existing usage in `ArbitrumBridge.tsx`, so the core mechanism is sound.

**Recommendation: approve with minor changes.** Nothing blocking, but a few things worth tightening.

### Should-fix

**1. `verifyCheckedIn` false-positive when `weekStart === 0`**

The standalone script guards this correctly (`last === weekStart && weekStart !== 0n`), but the in-component verification does not:

```ts
if (ws != null && lc != null) {
  return BigInt(lc as any) === BigInt(ws as any)
}
```

If both reads return `0` (uninitialized contract, fork/misconfigured RPC, or a brand-new hook), `0 === 0` reports a successful check-in that never happened — exactly the phantom-success bug this PR is killing. Add a `ws !== 0n` guard in both `fees.tsx` and `WeeklyRewardPool.tsx`, consistent with the script.

**2. Diagnostic script duplicates contract addresses from `const/config`**

`check-checkin-status.mjs` hardcodes `feeHook` / `vMooney` addresses per chain. They already live in `FEE_HOOK_ADDRESSES` (and vMOONEY config). Acceptable since it's a standalone `.mjs` not in the build, but it will silently drift if addresses change. A comment pointing back to `const/config` (or importing from it) would help.

### Nits

**3. Redundant `isCheckedIn` sources.** Three writers now: `refreshFeeData`, the new `useEffect([feeData])`, and `handleCheckIn`. The `useEffect([feeData])` duplicates logic already at the tail of `refreshFeeData`. Works fine, but consolidating to a single derived source reduces drift risk.

**4. Indentation in the de-nested block.** When `fetchStatus` was lifted into `refreshFeeData`, the `chains.map(...)` body kept its old (now over-indented) formatting. Harmless, but if Prettier is enforced in CI it'll flag — worth a `prettier --write` pass on both files.

**5. Timing/UX.** `verifyCheckedIn` does up to 3×800ms retries, the network-changed path adds 1s sleeps, and `switchToChain` waits up to 20s. A multi-chain check-in could feel long with no incremental progress. Not a correctness issue; a per-chain "verifying…" state could help later.

**6. `getAccountForChain` fallback to stale `account`.** The catch falls back to the original (stale-network) `account`. Reasonable as best-effort since on-chain verification will catch a genuine failure rather than report false success — just flagging it's intentional.

### Confirmed correct
- Retry-loop control flow: `break` on genuine errors skips verification; nonce errors fall through to verification; `retries--` at top prevents infinite loops.
- Success gated solely on on-chain state — no more trusting the tx promise or treating nonce collisions as success.
- Optimistic `setFeeData` + `finally` reconcile-after-1.5s pattern is reasonable for preventing no-op re-sends.

Net: good PR — ship after the `weekStart !== 0` guard (#1) and ideally a Prettier pass (#4).

pmoncada (approved):

PR #1345, opened 2026-06-01
Refresh profile to show new image on edit + success toast

pmoncada:
## Summary

When a user edits their profile (citizen or team), they previously saw their **old** profile after saving — the modal closed and the page only reloaded after a blind 30s timer, with no confirmation that the edit went through.

This change:
- **Adds a "Profile Editing Successful" toast** the moment the on-chain update is confirmed (plus an error toast on failure).
- **Refreshes the profile so the new image actually shows.** Instead of a blind 30s reload, we poll Tableland until the updated row (image, name, description) is indexed, then reload. The profile page is server-rendered from Tableland, so this guarantees the reload reflects the new image/details rather than the stale version — and it is usually faster than the old 30s wait.

Applies to both `CitizenMetadataModal` and `TeamMetadataModal` (same edit pattern).

## Details
- The poll is best-effort and time-bounded (20 attempts x 3s = up to 60s) and falls back to reloading anyway if indexing is slow, so the flow never gets stuck.
- The success toast fires and the modal closes immediately; the reload happens once the new data is live.

## Test plan
- [ ] Edit a citizen profile image, confirm the "Profile Editing Successful" toast appears, then the page refreshes showing the **new** image (not the old one).
- [ ] Edit only text fields (name/bio), confirm toast + refresh shows updated details.
- [ ] Same checks for a team profile.
- [ ] Reject/failed transaction shows an error toast and does not reload.

---

> [!NOTE]
> **Low Risk**
> UI-only post-transaction refresh and polling; no changes to auth, contracts, or data validation beyond existing Tableland query patterns.
> 
> **Overview**
> After a successful on-chain profile update, **citizen and team edit modals** now show a success toast, close the modal, **poll Tableland** until the row matches the new image, name, and description (instead of waiting 30 seconds and reloading blindly), then call `router.reload()`. Failed updates surface an **error toast**.
> 
> Supporting changes: **`unescapeQuotes`** in `cleanData.ts` so poll comparisons match Tableland-stored strings after `cleanData` escaping; **`waitForRow`** gains optional **`cacheBusting`** on the query API to avoid stale CDN responses. Table name constants (`CITIZEN_TABLE_NAMES` / `TEAM_TABLE_NAMES`) wire the SELECT used for that poll.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f9828899621e2dd4b73298addd5a51999f6e267. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


mmoncada1 (approved):

mmoncada1 (commented):
I like "Profile Edited Successfully" better for the toast the other one sounds off

PR #1357, opened 2026-06-04
fix: hide teams from non-citizens and constrain unlock banner width on profile page

mmoncada1:
## Summary

- **Teams hidden from non-citizens**: Added `(citizen || isOwner) &&` guard to the Teams section on the citizen profile page, matching the same guard already on the Governance section. Visitors who are not citizens and not the profile owner no longer see the Teams list.
- **Unlock banner width fixed**: Wrapped the "Unlock Full Profile" `Action` component in a `max-w-[1080px] mx-auto` container so it aligns with the rest of the page content instead of stretching full-width.

## Test plan

- [ ] View a citizen profile while **not logged in** — Teams section should not appear, Governance should not appear, and the "Unlock Full Profile" banner should be constrained to the same width as the header card.
- [ ] View a citizen profile while logged in as a **citizen** — Teams and Governance sections should appear as before.
- [ ] View your **own** citizen profile — all sections should appear as before.

### DIFF
--- a/ui/components/layout/ContentLayout.tsx
+++ b/ui/components/layout/ContentLayout.tsx
@@ -64,7 +64,7 @@ const ContentLayout: React.FC<ContentProps> = ({
                             `}
               style={{ maxWidth }}
             >
-              <div id="image-container" className="w-full h-full relative mb-10 z-10">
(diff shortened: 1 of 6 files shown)


pmoncada (approved):

mmoncada1:
Closing in favor of #1359 which isolates only the citizen profile page fixes.

PR #1358, opened 2026-06-04
feat: wire up MoonPay onramp ahead of Coinbase June 30 deprecation

mmoncada1:
## Summary

- **Coinbase onramp is 100% unchanged** — all existing flows, components, API routes, and hooks are identical to main. No user-facing behavior changes.
- Adds `MoonPayOnramp` and `MoonPayOnrampModal` components (new files) that use Privy's built-in `wallet.fund()` with MoonPay config.
- Generalizes the existing `useMoonPay` hook to support any chain (was previously hardcoded to `ETH_ARBITRUM`).
- No new environment variables or API keys required — MoonPay is managed entirely through the existing Privy app.

## Why

Coinbase's onramp API is deprecated June 30, 2026. This PR pre-wires MoonPay so the cutover on/after June 30 is a minimal diff (~10 lines, swapping imports in 3-4 consumer files).

## New files

| File | Purpose |
|------|---------|
| `ui/components/moonpay/MoonPayOnramp.tsx` | Drop-in replacement for `CBOnramp` — shows ETH amount, opens MoonPay widget, polls balance until sufficient, then fires `onBalanceSufficient()` callback |
| `ui/components/moonpay/MoonPayOnrampModal.tsx` | Modal wrapper for `MoonPayOnramp`, same interface as `CBOnrampModal` |
| `ui/lib/privy/hooks/useMoonPay.tsx` | Updated: chain-aware `getMoonPayCurrencyCode()` mapping + generalized `fund(amount?, chainId?)` |

## How to activate MoonPay when ready

1. In the Privy dashboard, ensure MoonPay is enabled as a funding provider for the app.
2. In each consumer file, swap the import and component name:
   - `CreateCitizen.tsx`: `CBOnrampModal` → `MoonPayOnrampModal`
   - `MissionContributeModal.tsx`: `CBOnramp` → `MoonPayOnramp`
   - `PrivyConnectWallet.tsx`: call `moonPayFund()` instead of `openCoinbaseOnramp()`

## Test plan

- [ ] Verify existing Coinbase onramp still works end-to-end (citizen creation, mission contribution, wallet fund button)
- [ ] Confirm no regressions in CI
- [ ] Before June 30: enable MoonPay in Privy dashboard and test the widget opens correctly


ui/lib/privy/hooks/useMoonPay.tsx
```diff
@@ -1,18 +1,47 @@
...
+    case 421614: // Arbitrum Sepolia
+      return 'ETH_ARBITRUM'
+    case 8453: // Base
+    case 84532: // Base Sepolia
+      return 'ETH_BASE'
+    case 137: // Polygon
+      return 'MATIC_POLYGON'
+    case 10: // Optimism — no ETH_OPTIMISM in MoonPay; fall back to Ethereum mainnet
+    case 11155420: // Optimism Sepolia
+      return 'ETH_ETHEREUM'
```
pmoncada: Correctness concern for Optimism (and Polygon below): this maps Optimism → `ETH_ETHEREUM`, so MoonPay would deliver ETH on **Ethereum mainnet**, not Optimism. In `MoonPayOnramp` the balance poll runs against `selectedChain` (Optimism), so it would never observe the funds and would hit the 30-minute timeout — even though the purchase succeeded.

Two ways the chain/currency can diverge here:
- Optimism: currency lands on mainnet, poll watches Optimism.
- Polygon (`MATIC_POLYGON`): the pre-filled `quoteCurrencyAmount` is computed as an ETH deficit, but the quote currency is MATIC, so the pre-fill amount is mislabeled.

Suggest restricting the supported chains to those where the native asset actually matches (Arbitrum / Base / Ethereum), and disabling MoonPay or showing a clear "funds arrive on Ethereum mainnet" warning for Optimism, rather than silently mapping it.

ui/lib/privy/hooks/useMoonPay.tsx
```diff
@@ -1,18 +1,47 @@
...
+  async function fund(amount?: number, chainId?: number) {
+    const currencyCode = getMoonPayCurrencyCode(chainId)
 
     const fundWalletConfig: MoonpayConfig = {
-      currencyCode: 'ETH_ARBITRUM',
+      currencyCode,
       paymentMethod: 'credit_debit_card',
-      quoteCurrencyAmount: roundedCurrencyAmount,
+      ...(amount && amount > 0
+        ? { quoteCurrencyAmount: Math.round(amount * 1000) / 1000 }
```
pmoncada: Minor: `quoteCurrencyAmount` is rounded to 3 decimals here, but `MoonPayOnramp` displays the "Amount needed" with `toFixed(4)`. For small ETH deficits the pre-filled amount can be up to 0.0005 ETH below what the UI shows, which could leave the purchase slightly short and force the user to re-onramp. Consider rounding *up* (ceil) to the displayed precision so the pre-fill always covers the deficit.

ui/components/moonpay/MoonPayOnramp.tsx
```diff
@@ -0,0 +1,382 @@
...
+}: MoonPayOnrampProps) {
+  const shellWidthClass = fullWidth ? 'w-full' : 'w-full max-w-md mx-auto'
+  const fund = useMoonPay()
+
+  const [fundingState, setFundingState] = useState<FundingState>('idle')
+  const [error, setError] = useState<string | null>(null)
+  const [pollCount, setPollCount] = useState(0)
+
+  // Derive the chain's native token symbol for display
+  const nativeSymbol: string = useMemo(
```
pmoncada, line 77: `nativeSymbol` is derived from `selectedChain.nativeCurrency.symbol` for display, but the asset MoonPay actually purchases is determined independently by `getMoonPayCurrencyCode(chainId)` in the hook. These can disagree (e.g. Optimism shows "ETH" on the Optimism network here, but the hook funds ETH on Ethereum mainnet). Display label and funding currency should come from the same source of truth so the UI can't promise something the funding flow won't deliver.

ui/components/moonpay/MoonPayOnramp.tsx
```diff
@@ -0,0 +1,382 @@
...
+      }
+
+      // Opens the MoonPay widget — resolves when widget closes
+      await fund(ethAmount > 0 ? ethAmount : undefined, selectedChain?.id)
+
+      setFundingState('waiting')
+      onPurchaseSubmitted?.()
+
+      if (checkBalanceSufficient) {
+        startPolling()
```
pmoncada, line 214: The whole polling/auto-proceed flow hinges on `wallet.fund()` resolving when the user *closes* the MoonPay widget (not when it merely opens). If Privy resolves this promise earlier, `onPurchaseSubmitted()` and `startPolling()` fire while the user is still mid-purchase. Worth confirming against the Privy `fund()` semantics for the installed version before cutover, and noting it in a comment here.

ui/components/privy/PrivyConnectWallet.tsx
```diff
@@ -680,10 +680,8 @@ export function PrivyConnectWallet({ citizenContract, type }: PrivyConnectWallet
     }
 
     try {
-      // Generate session token
       const token = await generateSessionToken()
```
pmoncada: This file's net diff is only incidental: the add-then-revert across commits left behind deleted inline comments (and a reformatted "} else if") with no functional change. Recommend restoring this file to `main` so the PR stays purely additive — it keeps the Coinbase path byte-identical and avoids needless merge-conflict surface. Same applies to CreateTeam.tsx and GuestActions.tsx, where only the viemChains import position changed.

pmoncada (commented):
Reviewed the MoonPay pre-wiring — left detailed inline comments. Overall this is a clean, low-risk approach.

**What's good**
- The net diff leaves the Coinbase onramp fully intact; the new `MoonPayOnramp`/`MoonPayOnrampModal` are dormant and nothing imports them yet, so there's no user-facing behavior change today.
- `useMoonPay` is sensibly generalized to be chain-aware.
- The Copilot-pass fixes (recursive `setTimeout` + in-flight guard instead of `setInterval`, `stopPolling()` on modal close, and `nativeCurrency.symbol` for display) are real improvements.

**Worth addressing before the cutover**
1. **Currency-vs-chain correctness (Optimism/Polygon).** `getMoonPayCurrencyCode` maps Optimism → `ETH_ETHEREUM` (mainnet) and Polygon → `MATIC_POLYGON`, but the widget amount and the balance poll are computed against `selectedChain`. On those chains funds land on a different network/asset than the poll watches, so it never auto-proceeds and hits the 30-min timeout despite a successful purchase. Restrict the supported set or make display + funding share one source of truth.
2. **Incidental diff noise.** `PrivyConnectWallet.tsx`, `CreateTeam.tsx`, and `GuestActions.tsx` only contain round-trip artifacts (deleted comments + reordered `viemChains` import). Reverting them to `main` keeps the PR purely additive and reduces merge-conflict risk.
3. **Untested path.** Since nothing exercises the new components yet, the cutover day would be their first real run. A minimal render/smoke test would de-risk that.
4. **`fund()` resolution timing.** The auto-proceed flow assumes Privy's `wallet.fund()` resolves on widget *close*; worth confirming for the installed Privy version.

None of these block merging given the components are dormant, but #1 should be sorted before flipping the switch.

pmoncada (approved):

pmoncada:
**Review: Request changes 🔴 — mislabeled scope and the build is broken**

The description says *"Coinbase onramp is 100% unchanged… adds MoonPay components (new files)… no new env vars… ~10-line cutover."* The actual diff (`+4330/-846`, 36 files) is a **major refactor**:

1. **`@privy-io/react-auth` v1.92 → v3.29** (major bump) and removal of `@privy-io/wagmi-connector`. Touches the whole auth layer, including a **193-line rewrite of `PrivyConnectWallet.tsx`** plus edits to providers, `useSafe`, `useNativeBalance`, `useUniswapV4`.
2. **New headless Coinbase onramp:** `CBHeadlessOnramp.tsx` (+625), `FundOnramp*`, `OnrampReturnHandler`, and **new API routes** `create-onramp-order.ts` (+258) and `region.ts` (CDP credentials, geo/region detection, Upstash Redis). So Coinbase is very much *not* unchanged.
3. Build hacks to survive the upgrade: new `.yarnrc` (`ignore-engines`), CI changed to `rm -rf node_modules && yarn install --frozen-lockfile`, cache-key bump, ~11 webpack `resolve.alias … false` entries (Solana/Stripe/Farcaster), CSP changes for MoonPay/Meld/Google Pay.
4. **"No new env vars" is inaccurate** — `create-onramp-order.ts` reads `UPSTASH_REDIS_URL/TOKEN`, CDP credentials, and `NEXT_PUBLIC_MOCK_ONRAMP`.

**CI is red:** all 4 `component-tests` shards fail to compile with:

```
ERROR in ./node_modules/@scure/bip32/lib/esm/index.js
Module not found: Package path ./legacy is not exported from .../@scure/bip32/node_modules/@noble/hashes
```

The `yarn.lock` churn (+1962/-593) from the Privy v3 upgrade left an inconsistent nested `@noble/hashes`. The `rm -rf node_modules` change fixed the `build` job, but the Cypress component webpack still resolves the broken nested version.

Recommendations:
- Fix the build first (likely a `resolutions` pin for `@noble/hashes` / `@scure/bip32`).
- Update the PR description to reflect the real scope (Privy v3, headless Coinbase, new API routes, new env vars).
- Full manual QA of **all** auth flows (login, embedded wallet, Safe, balance hooks, swap, bridge) given Privy v3 breaking changes — not just the onramp.
- Ideally split the Privy v3 upgrade from the MoonPay wiring into separate PRs.

ui/lib/privy/hooks/useMoonPay.tsx
```diff
@@ -1,26 +1,60 @@
...
     }
 
-    return wallets[selectedWallet].fund({ config: fundWalletConfig })
+    return fund({
+      // Default the fiat picker to USD but allow the user's local currency too.
+      source: { defaultAsset: 'usd' },
+      destination: {
+        asset: 'eth',
+        chain: `eip155:${chainId}`,
+        address: wallet.address,
```
pmoncada, line 50: **MoonPay funds the selected wallet, ignoring the `address` recipient prop (important).** `MoonPayOnramp` accepts and validates an explicit `address` prop, but the destination passed to `fund()` is `wallets[selectedWallet].address` from the hook — the prop is never used. If `selectedWallet` ever points to a different wallet than the displayed `address`, funds silently route to the wrong address. Fix: pass the recipient `address` into `useMoonPay`/`fund()`, or assert `wallet.address.toLowerCase() === address.toLowerCase()` and throw on mismatch.

ui/lib/privy/hooks/useMoonPay.tsx
```diff
@@ -1,26 +1,60 @@
...
+      },
+      environment: 'production',
```
pmoncada, line 52: **`environment: 'production'` hardcoded despite testnet chain IDs + mock flag (important).** `SUPPORTED_MOONPAY_CHAIN_IDS` includes Sepolia / Arbitrum Sepolia / Base Sepolia, and the codebase has a `NEXT_PUBLIC_MOCK_ONRAMP` sandbox concept, but the onramp environment is always `'production'`. Test/staging funding on testnets can't be delivered by a production provider. Fix: derive `environment` from `NEXT_PUBLIC_MOCK_ONRAMP`/chain ('sandbox' for testnets).

ui/lib/privy/hooks/useMoonPay.tsx
```diff
@@ -1,26 +1,60 @@
...
+import { useFiatOnramp, useWallets } from '@privy-io/react-auth'
 import { useContext } from 'react'
 import PrivyWalletContext from '../privy-wallet-context'
 
+/**
+ * Chain IDs where Privy's in-app fiat onramp can deliver the chain's native
+ * token (ETH) directly. These are all ETH-native networks, so the destination
+ * asset is always "eth".
+ */
+export const SUPPORTED_MOONPAY_CHAIN_IDS = new Set([
```
pmoncada, line 10: **`SUPPORTED_MOONPAY_CHAIN_IDS` is defined but never enforced (important).** `fundWallet` only checks `if (!chainId)` — it never validates membership in this set, so a call with an unsupported chain still attempts the onramp and fails opaquely. Fix: `if (!SUPPORTED_MOONPAY_CHAIN_IDS.has(chainId)) throw new Error(...)`.

ui/components/onramp/FundOnrampModal.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+export const FundOnrampModal: React.FC<FundOnrampModalProps> = ({
+  enabled,
+  setEnabled,
+  address,
+  selectedChain,
+  ethAmount,
+  onExit,
+  isWaitingForGasEstimate = false,
+  defaultProvider = 'moonpay',
```
pmoncada, line 51: **Forcing `defaultProvider='moonpay'` disables US→Coinbase auto-detection and shows misleading copy (important).** The modal always passes `defaultProvider`, so `FundOnramp`'s region effect early-returns and `isUS` stays `null` forever. Every onboarding user defaults to MoonPay regardless of region, and a US user who manually selects Coinbase sees the non-US warning 'Only choose Coinbase if you already have a Coinbase account' even though the US Headless flow is account-less Apple Pay. Fix: let region detection run even when a default is set (use `defaultProvider` only as the initial value), or gate the warning on `isUS === false`.

ui/pages/api/coinbase/region.ts
```diff
@@ -0,0 +1,46 @@
...
+  url: process.env.UPSTASH_REDIS_URL!,
+  token: process.env.UPSTASH_REDIS_TOKEN!,
+})
+
+/**
+ * Lightweight geo lookup so the UI can choose the right Coinbase onramp flow:
+ * the Headless (Apple Pay / Google Pay) checkout is US-only, while the hosted
+ * redirect remains the fallback for everyone else.
+ */
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
```
pmoncada, line 20: **`region.ts` is an unauthenticated public endpoint that instantiates Redis (important/nit).** Unlike `create-onramp-order` (wrapped in `authMiddleware`), this route has no auth or rate limiting and calls `detectUserState(req, redis)` (which caches IP→state in Redis). It's trivially spammable, amplifying Redis cost. Fix: add the `authMiddleware`/`secureHeaders` wrapper used elsewhere, or at least rate-limit; consider in-memory caching of the geo result.

ui/components/coinbase/CBHeadlessOnramp.tsx
```diff
@@ -0,0 +1,625 @@
...
+            </div>
+          )}
+          <div className="bg-black/20 rounded-lg border border-white/10 overflow-hidden">
+            {/* Safari renders just the Apple Pay button (short); other browsers
+                render a full QR-code screen that needs much more height. */}
+            <iframe
+              title="Coinbase payment"
+              src={paymentLinkUrl}
+              allow="payment *; camera *; microphone *; accelerometer *; gyroscope *; magnetometer *"
+              sandbox="allow-scripts allow-same-origin"
```
pmoncada, line 439: **iframe uses `allow-scripts allow-same-origin` together (nit).** Combining these two tokens negates the sandbox's isolation. It's acceptable here because `src` is a Coinbase-returned `paymentLink.url` (trusted) and Apple Pay needs `allow="payment"`, but worth a comment noting the deliberate trade-off and confirming the URL is always validated as a Coinbase origin before mounting.

pmoncada (commented):
**Review summary — MoonPay onramp**

Note: this routes fiat through Privy's in-app `useFiatOnramp()` (MoonPay/Coinbase/Meld by region) rather than a direct MoonPay integration — so there are **no MoonPay secrets, URL signing, or webhooks to verify**, and delivery is confirmed by on-chain balance polling (safe). I verified no `NEXT_PUBLIC_*` secret was introduced and Coinbase CDP creds stay server-side. The substantive findings are correctness/UX: the recipient `address` prop is ignored in favor of the selected wallet, `environment` is hardcoded to `'production'` despite testnet chain IDs, `SUPPORTED_MOONPAY_CHAIN_IDS` is never enforced, forcing `defaultProvider='moonpay'` disables US→Coinbase auto-detection, and `region.ts` is an unauthenticated Redis-backed endpoint. See inline comments.

PR #1362, opened 2026-06-05
Skip cookie consent banner for US users

mmoncada1:
## Summary
- US visitors don't legally need to make a cookie/consent decision, so the consent banner is now hidden for them and analytics consent is auto-granted.
- Adds `GET /api/geo/country`, which returns the requester's country from Vercel/Cloudflare geo headers (`getCountryFromHeaders`) for the client to read.
- `CookieBanner` now resolves consent before rendering: respects any stored decision, auto-grants + hides for `US`, and otherwise shows the banner. Any geo lookup failure falls back to showing the banner (privacy-protective default).

## Test plan
- [ ] Non-US / unknown geo: banner renders; Accept/Decline/X update `localStorage` + `gtag` as before.
- [ ] US geo (`/api/geo/country` -> `{ country: 'US' }`): banner does not render, `gtag('consent','update',{analytics_storage:'granted'})` fires, `cookie_consent` stored as `true`.
- [ ] Existing stored consent is respected regardless of country.
- [ ] Cypress: `cypress/integration/layout/cookie-banner.cy.tsx` (geo endpoint stubbed; added US case).

### DIFF
--- a/ui/components/layout/CookieBanner.tsx
+++ b/ui/components/layout/CookieBanner.tsx
@@ -4,29 +4,56 @@ import { useEffect, useState } from 'react'
 
 export default function CookieBanner() {
   const [cookieConsent, setCookieConsent] = useState<any>(false)
-  const [hasLoadedLocalStorage, setHasLoadedLocalStorage] = useState(false)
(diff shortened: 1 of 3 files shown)


pmoncada (approved):
test it before merging

PR #1367, opened 2026-06-08
fix: cleaner buy, bridge, and lock page UX (to main)

mmoncada1:
## Summary

Re-targets the changes from #1361 directly at `main`. The original PR was stacked on `fix/citizen-profile-page` and merged there instead of `main`, so these fixes never reached `main`. This branch cherry-picks the same 5 commits onto `main` (identical diff, no conflicts).

Fixes UX issues across the **buy (get-mooney)**, **bridge**, and **lock** pages.

### Buy / Get MOONEY (`components/uniswap/NativeToMooney.tsx`)
- MAX button reserves gas before filling the amount so the route quote no longer fails.
- Removed double max-clamping that broke the cursor/deletion when typing over balance; button now shows "Not Enough …" and is disabled.
- Confetti fires on a successful swap.

### Bridge (`components/bridge/ArbitrumBridge.tsx`)
- Same clamping/cursor fix; button shows "Not Enough …" / "Enter Amount" and is disabled instead of silently snapping the value down.

### Lock (`pages/lock.tsx`)
- Explicit "Add MOONEY" / "Extend Lock" toggle for existing locks (since `increaseLock` only does one operation per tx).
- Fixed max amount (locked + wallet balance) and guarded `parseEther` against intermediate input.

### Sign-in (`components/privy/PrivyWeb3Button.tsx`)
- Page-specific sign-in labels for signed-out users.

## Test plan
- Buy: MAX with low ETH balance → route resolves and swap is enabled.
- Buy: type over balance → cursor/deletion normal; button shows "Not Enough ETH"; swap on success fires confetti.
- Bridge: type over balance → cursor/deletion normal; button shows "Not Enough …".
- Lock (new): amount + duration both required.
- Lock (existing): toggle between "Add MOONEY" and "Extend Lock"; each runs only its own transaction.

Supersedes #1361.

### DIFF
--- a/ui/components/bridge/ArbitrumBridge.tsx
+++ b/ui/components/bridge/ArbitrumBridge.tsx
@@ -264,6 +264,11 @@ export default function ArbitrumBridge() {
     setSelectedChain(ethereum)
   }, [setSelectedChain])
 
+  const numAmount = parseFloat(String(amount)) || 0
+  const isOverBalance =
+    balance !== undefined && numAmount > 0 && numAmount > Number(balance)
+  const tokenLabel = inputToken === 'eth' ? 'ETH' : 'MOONEY'
+
   return (
     <div className="w-full mt-3 sm:mt-4">
       <div className="text-sm font-RobotoMono rounded-xl sm:rounded-2xl animate-fadeIn p-3 sm:p-4 flex flex-col bg-gradient-to-br from-gray-900 via-blue-900/30 to-purple-900/20 backdrop-blur-xl border border-white/10 shadow-2xl text-white">
@@ -305,7 +310,6 @@ export default function ArbitrumBridge() {
                     className="text-white bg-transparent text-xl sm:text-2xl font-RobotoMono placeholder-gray-500 focus:outline-none w-full min-w-0 border-0 !p-0 min-h-[28px] tabular-nums"
                     bare
                     value={amount}
-                    max={balance ? Number(balance) : undefined}
                     onChange={(e) => {
                       let value = e.target.value
                       value = value.replace(/[^0-9.]/g, '')
@@ -314,9 +318,6 @@ export default function ArbitrumBridge() {
                         value = parts[0] + '.' + parts.slice(1).join('')
                       }
                       if (parseFloat(value) < 0) value = '0'
-                      if (balance && parseFloat(value) > balance) {
-                        value = String(balance)
-                      }
                       if (value.startsWith('0') && value.length > 1 && value[1] !== '.') {
                         value = value.substring(1)
                       }
@@ -399,8 +400,14 @@ export default function ArbitrumBridge() {
               skipNetworkCheck={skipNetworkCheck}
               requiredChain={ethereum}
               className="w-full bg-gradient-to-r from-blue-500 to-purple-600 hover:from-blue-600 hover:to-purple-700 text-white py-3 sm:py-4 px-4 rounded-xl text-base sm:text-lg font-semibold transition-all duration-200 transform hover:scale-[1.02] shadow-lg disabled:opacity-50 disabled:cursor-not-allowed disabled:hover:scale-100 disabled:from-gray-500 disabled:to-gray-600"
-              label="Bridge to Arbitrum"
-              isDisabled={!isEOA}
+              label={
+                numAmount === 0
+                  ? 'Enter Amount'
+                  : isOverBalance
+                  ? `Not Enough ${tokenLabel}`
+                  : 'Bridge to Arbitrum'
+              }
+              isDisabled={!isEOA || numAmount === 0 || isOverBalance}
               action={async () => {
                 try {
                   if (inputToken === 'eth') {
--- a/ui/components/bridge/ArbitrumBridge.tsx
+++ b/ui/components/bridge/ArbitrumBridge.tsx
@@ -11,6 +11,7 @@ import { useActiveAccount, useActiveWallet } from 'thirdweb/react'
 import { EIP1193 } from 'thirdweb/wallets'
 import PrivyWalletContext from '../../lib/privy/privy-wallet-context'
 import { arbitrum, ethereum } from '@/lib/rpc/chains'
+import { useGasPrice } from '@/lib/rpc/useGasPrice'
 import { getChainSlug } from '@/lib/thirdweb/chain'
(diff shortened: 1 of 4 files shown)


pmoncada:
**Review: Approve after a manual pass 🟡**

Re-targets #1361 onto `main` (the original merged into the wrong base). `build` + `component-tests` pass.

- Lock page: explicit **Add MOONEY / Extend Lock** toggle (since `increaseLock` does one op per tx), `safeParseEther` guard against intermediate input, `isOverBalance` disabling. Logic reads correctly.
- Buy/Bridge: MAX reserves gas before quoting, removes the double-clamping that broke cursor/deletion, disabled "Not Enough…" state, confetti on success.
- Mostly UX with money-adjacent inputs — worth the manual QA checklist in the PR (especially MAX-with-low-ETH, and the existing-lock toggle running only its own tx).

CI note: `e2e-tests` failures appear across all open PRs (pre-existing/flaky on `main`).

ui/components/uniswap/NativeToMooney.tsx
```diff
@@ -38,9 +56,13 @@ export default function NativeToMooney({ selectedChain }: any) {
   const { effectiveGasPrice } = useGasPrice(selectedChain)
   const { ethPrice } = useETHPrice(1, 'ETH_TO_USD')
 
+  const isOverBalance =
```
pmoncada, line 70: **Typing the exact full ETH balance still leaves no gas (nit).** The gas-reserving MAX is a nice fix, but `isOverBalance` only blocks `amount > balance`. A user who manually types their full balance passes the gate and the swap fails on gas. For consistency with the new MAX, consider subtracting the same gas reserve in `isOverBalance` so the button reads 'Not Enough ETH' instead of letting it fail.

ui/pages/lock.tsx
```diff
@@ -204,6 +223,63 @@ export default function Lock() {
...
+    ? hasLock && VMOONEYLock
+      ? parseFloat(
+          ethers.utils.formatEther(
+            BigNumber.from(VMOONEYLock[0]).add(MOONEYBalance)
+          )
+        )
+      : parseFloat(ethers.utils.formatEther(MOONEYBalance.toString()))
+    : undefined
+
+  const isOverBalance =
```
pmoncada, line 246: **Lock over-balance check loses precision via `parseFloat` (nit).** Both sides come from `parseFloat(formatEther(...))`. For very large MOONEY balances `parseFloat` drops low-order wei, so `isOverBalance` can read `false` for an amount a few wei over. Harmless in practice (the contract enforces the true limit), just flagging that the UI gate is approximate.

ui/components/bridge/ArbitrumBridge.tsx
```diff
@@ -264,6 +264,11 @@ export default function ArbitrumBridge() {
     setSelectedChain(ethereum)
   }, [setSelectedChain])
 
+  const numAmount = parseFloat(String(amount)) || 0
+  const isOverBalance =
+    balance !== undefined && numAmount > 0 && numAmount > Number(balance)
```
pmoncada: **ETH over-balance gate ignores gas → guaranteed revert (important).** For ETH, `balance` is the full L1 ETH balance, so a user entering (near-)full balance passes this gate (`numAmount > Number(balance)` is false) and the button stays enabled as 'Bridge to Arbitrum' — then `ethBridger.deposit` reverts because no ETH is left for L1 gas. Consider factoring a gas reserve into `isOverBalance` for the ETH case so it reads 'Not Enough ETH' instead of failing on-chain.

pmoncada: **MOONEY balance is rounded to 2 decimals before driving this gate (nit).** For MOONEY, `balance` resolves to `ethMooneyBalance`, which is computed as `(balance.toString() / 10**18).toFixed(2)` (line 199) and can round *up* (e.g. 10.005 → '10.01'). Since this new over-balance gate compares against that rounded value, a user can pass the check with slightly more MOONEY than they actually hold and the bridge `transferFrom` reverts. Consider comparing against a non-rounded balance.

pmoncada (commented):
**Review summary — buy/bridge/lock UX cleanup**

Solid UX cleanup. The new "Not Enough X"/"Enter Amount" button states, `signInLabel` support, gas-reserving MAX on the buy page, and `safeParseEther` crash-fix are all well done. Verified no build-break (`canvas-confetti` + types are declared deps) and the new MAX BigInt math is type-safe (`effectiveGasPrice` is a real `bigint`). Main item: the new ETH over-balance gate in the bridge ignores L1 gas, so entering (near-)full ETH balance passes the gate and `deposit` reverts — inconsistent with the gas-aware buy page. See inline comments (all important/nit, no fund-loss bugs).

pmoncada (approved):

ui/components/uniswap/NativeToMooney.tsx
mmoncada1: Fixed in 59e30f40c. `isOverBalance` now subtracts the same gas reserve used by MAX (`spendableBalance = balance - gasReserveEth`), so typing the full ETH balance reads 'Not Enough ETH' instead of failing on gas.

ui/pages/lock.tsx
mmoncada1: Fixed in 59e30f40c. The lock over-balance gate now compares in wei via BigNumber (`safeParseEther(lockAmount).gt(maxAmountBN)`) instead of `parseFloat(formatEther(...))`, so no low-order wei are dropped.

ui/components/bridge/ArbitrumBridge.tsx
mmoncada1: Fixed in 59e30f40c. The ETH over-balance gate (and MAX) now reserve L1 gas via `useGasPrice(ethereum)` (`effectiveGasPrice * 200000`, fallback 0.002 ETH), so a near-full balance reads 'Not Enough ETH' rather than reverting in `deposit`. MOONEY keeps no reserve since its gas is paid separately in ETH.

mmoncada1: Fixed in 59e30f40c. `ethMooneyBalance`/`arbMooneyBalance` now store the non-rounded `formatEther` value instead of `.toFixed(2)`, so the over-balance gate compares against the true balance (display still rounds via `toFixed(5)`).

PR #1372, opened 2026-06-10
feat(contributions): Discord notification for new Community Circle submissions

mmoncada1:
## Summary

Community Circle contributions are submitted through an **external Google Form** that writes rows to a Google Sheet. The app only ever *reads* that sheet (`getSheetContributions`) to render the feed — there is **no server-side hook on submit**, so the previous Discord notification (an external Form/Sheet automation) is not something this repo could fire. This PR rebuilds the notification entirely in-repo as a scheduled poller.

> Note: an earlier commit on this branch patched `sendDiscordMessage` (the Coordinape `ContributionEditor` path). That flow is **not** used for contributions, so that change was reverted in this branch.

### How it works
- **`ui/lib/contributions/notifyNewContributions.ts`** — reads the contributions sheet, diffs it against rows already announced (tracked in **Upstash Redis**), and posts each new row to the network-notifications Discord channel (`GENERAL_CHANNEL_ID` on mainnet / `TEST_CHANNEL_ID` otherwise) via `DISCORD_BOT_TOKEN`.
  - **First-run seeding:** on the very first run it records all existing rows as "seen" *without* posting, so enabling this doesn't dump the entire historical backlog into Discord.
  - **Idempotent + safe:** rows are only marked seen after a successful Discord post (failures retry next run), capped at 8 posts/run to avoid rate limits.
- **`ui/pages/api/cron/contribution-notifications.ts`** — secret-protected endpoint (`CRON_SECRET` via `Authorization: Bearer`, `x-cron-secret`, or `?secret=`), mirroring the existing townhall cron auth.
- **`.github/workflows/contribution-notifications.yml`** — GitHub Actions cron that polls the endpoint every 10 minutes (matches the repo's existing scheduled-job pattern).

### Required configuration
- `CONTRIBUTIONS_SHEET_CSV_URL`, `UPSTASH_REDIS_URL`, `UPSTASH_REDIS_TOKEN`, `DISCORD_BOT_TOKEN` (env on the deployed site)
- Repo secrets for the workflow: `CRON_SECRET` and optionally `MOONDAO_SITE_URL` (defaults to `https://moondao.com`)

## Test plan
- [ ] Set the env vars above; trigger the workflow manually (`workflow_dispatch`) once — confirm the **first run seeds silently** (no Discord posts) and returns `{ seeded: true }`.
- [ ] Submit a test contribution via the Google Form; on the next poll confirm a Discord message posts to the network-notifications channel.
- [ ] Re-run the cron and confirm the same contribution is **not** re-posted (dedup works).
- [ ] Verify a missing/invalid `CRON_SECRET` returns 401.

### DIFF
--- a/ui/lib/contributions/notifyNewContributions.ts
+++ b/ui/lib/contributions/notifyNewContributions.ts
@@ -28,6 +28,15 @@ import {
 const NOTIFIED_SET_KEY = 'contributions:notified'
 const SEEDED_FLAG_KEY = 'contributions:notified:seeded'
 
+// Per-run lock so overlapping cron invocations can't both read the seen-set and
+// double-post the same row (the gap between SMEMBERS and SADD). The release is a
+// compare-and-delete keyed on a unique token, so a slow run can never delete a
+// lock that a later run has already acquired.
+const LOCK_KEY = 'contributions:notify:lock'
+const LOCK_TTL_SECONDS = 120
+const RELEASE_LOCK_SCRIPT =
+  'if redis.call("get", KEYS[1]) == ARGV[1] then return redis.call("del", KEYS[1]) else return 0 end'
+
 // Cap posts per run so a sudden burst (or a misconfiguration) can't spam the
 // channel or trip Discord's rate limiter. Anything beyond this is picked up on
 // the next poll, since rows are only marked seen after a successful post.
@@ -49,6 +58,8 @@ export type NotifyResult = {
   total?: number
   /** New rows that remain unposted (deferred to the next run / failed). */
   pending?: number
+  /** Set when another run held the lock and this invocation was skipped. */
+  locked?: boolean
   /** Populated when ok is false. */
   reason?: string
 }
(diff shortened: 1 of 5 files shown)


pmoncada:
**Review: Approve with minor follow-ups 🟡**

Architecture is sound: poll the sheet, diff against an Upstash Redis set of seen row-hashes, seed silently on first run, cap 8 posts/run, and mark-seen only after a successful Discord post.

Things to flag:
- **Open endpoint when secret unset:** `/api/cron/contribution-notifications` is unauthenticated if `CRON_SECRET` is unset, and accepts `GET` too. It mirrors the existing townhall cron, and the work is idempotent/rate-limited, so impact is low — but please ensure `CRON_SECRET` is set in prod.
- **No run-lock:** overlapping cron runs could double-post a row (gap between `smembers` read and `sadd`). Unlikely at a 10-min interval, but a per-run Redis lock would make it airtight.
- Row identity is `sha1(timestamp|wallet|description)` — editing a submission's description later would re-post it. Acceptable.

ui/lib/contributions/notifyNewContributions.ts
```diff
@@ -0,0 +1,213 @@
...
+  try {
+    const resp = await fetch(
+      `https://discord.com/api/v10/channels/${NOTIFICATION_CHANNEL_ID}/messages`,
+      {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          Authorization: `Bot ${token}`,
+        },
+        body: JSON.stringify({ content }),
```
pmoncada: **Discord mention injection (blocking).** `buildContent` interpolates `c.name`, `c.area`, `c.description`, `c.links` — all fields a member of the public types into a Google Form (no sanitization). Posting as top-level `content` means Discord parses `@everyone`, `@here`, and `<@&roleId>`/`<@userId>` and pings them, so anyone who can submit the form can mass-ping the channel. (`mission/contribution-notification.ts` avoids this by using `embeds[].description`.) Fix: add `allowed_mentions: { parse: [] }` to the request body.

ui/lib/contributions/notifyNewContributions.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  try {
+    const contributions = await getSheetContributions()
+    // `getSheetContributions` returns newest-first; post oldest-first so the
+    // Discord feed reads chronologically.
+    const ordered = [...contributions].reverse()
+    const keys = ordered.map(contributionKey)
+
+    const seeded = await redis.get(SEEDED_FLAG_KEY)
```
pmoncada, line 194: **First-run seeding flips the 'seeded' flag even on a failed sheet fetch (important).** `getSheetContributions()` returns `[]` both when the sheet is genuinely empty and when the fetch transiently fails / the CSV URL is momentarily unset. If the first run hits such a failure, `keys.length === 0`, nothing is added to the seen-set, yet `SEEDED_FLAG_KEY` is still set — so the next successful run dumps the entire historical backlog (8/run). Fix: only set the seeded flag when `keys.length > 0`.

ui/lib/contributions/notifyNewContributions.ts
```diff
@@ -0,0 +1,213 @@
...
+      return { ok: true, seeded: false, posted: 0, total: 0 }
+    }
+
+    const seen = new Set(await redis.smembers(NOTIFIED_SET_KEY))
+    const newOnes: { c: Contribution; key: string }[] = []
+    for (let i = 0; i < ordered.length; i++) {
+      if (!seen.has(keys[i])) newOnes.push({ c: ordered[i], key: keys[i] })
+    }
+
+    let posted = 0
```
pmoncada, line 218: **One un-postable row permanently blocks the queue (important).** `newOnes` is oldest-first and `if (!ok) break` stops without marking the failing row seen, so if the oldest row is permanently rejected by Discord, every run retries it first, breaks, and never progresses. Realistic trigger: total message length > Discord's 2000-char limit (`area`, `who`/`name`, and `links` aren't length-capped). Fix: cap total content length in `buildContent`, and/or skip-and-mark-seen (or count failures) instead of an unconditional `break`.

ui/pages/api/cron/contribution-notifications.ts
```diff
@@ -0,0 +1,47 @@
+import type { NextApiRequest, NextApiResponse } from 'next'
```
pmoncada, line 1: **No request timeout on fetches; cron route sets no `maxDuration` (nit).** `postToDiscord` and the sheet fetch use `fetch` with no `AbortController`/timeout, and this route doesn't export `config = { maxDuration }` like `mission/contribution-notification.ts` does. With `MAX_PER_RUN=8` + 400ms sleeps + latency, a slow run can exceed the default function limit and be killed mid-loop. Impact is bounded (dedup + lock TTL), but an explicit fetch timeout and `maxDuration` would make behavior predictable. Also note: when `CRON_SECRET` is unset the route is publicly triggerable, and accepting the secret via `?secret=` risks leaking it into access logs.

pmoncada (commented):
**Review summary — Community Circle Discord notifications**

Sound architecture (poll-and-diff with Redis dedup, per-run lock, seed-on-first-run, per-run caps), correctly mirroring the existing townhall cron patterns. Three issues to address before merge: (1) **blocking** — form input is posted as a raw Discord `content` with no `allowed_mentions`, so any public submitter can `@everyone`/role-ping the channel; (2) the first-run seed can set its "seeded" flag on a *failed* sheet fetch, later dumping the whole backlog; and (3) a single un-postable row (e.g. >2000 chars) permanently stalls the queue via the unconditional `break`. See inline comments.

pmoncada (approved):

pmoncada:
test in prod @mmoncada1

PR #1373, opened 2026-06-10
fix(quests): gate single-quest Claim on server's authoritative eligibility

mmoncada1:
## Summary

Follow-up hardening for the quest claim-button report (fresh citizen seeing "Claim" on Join a Team / Citizen Profile). Investigation showed the current code fails closed in every error path, but the Claim gating for single (non-staged) quests had real latent gaps:

- **Claim now gates on the proof route's authoritative `eligible` flag** instead of the client-side `metric >= 1` heuristic. `eligible` applies the per-verifier threshold server-side and is the same field `useClaimableQuestsCount` uses for the dashboard "N quests to claim" badge — so the badge and quest cards can no longer disagree. This also fixes any future single quest with a threshold above 1, which would previously have shown Claim too early.
- **Stale-response guard** in the eligibility effect: an in-flight eligibility fetch for a previous wallet/quest can no longer set state after an account switch (out-of-order resolution).
- **Eligibility refreshes after recovery actions**: the "Link GitHub" and "Refresh" error buttons previously fetched the metric and discarded the result, leaving the Claim state stale until remount. They now go through a shared `refreshSingleQuestEligibility` helper.

No backend changes; the POST claim path already re-verifies eligibility server-side before signing, so payouts were never at risk.

## Test plan

- [ ] Fresh citizen (empty profile, no team hats): Join a Team and Citizen Profile show "Not started", no Claim button, badge count 0
- [ ] Citizen with completed profile (description + location/discord/twitter/website): Citizen Profile shows Claim and the badge counts it
- [ ] Citizen wearing a team hat in the MoonDAO tree: Join a Team shows Claim
- [ ] Link GitHub from the Submit an Issue card: Claim state updates without a page reload
- [ ] Switch wallets while quest cards are loading: no Claim button leaks from the previous wallet's responses

### DIFF
--- a/ui/components/xp/Quest.tsx
+++ b/ui/components/xp/Quest.tsx
@@ -212,10 +212,19 @@ export default function Quest({
     [xpManagerContract, userAddress, quest.verifier.verifierId]
   )
 
-  // Function to fetch user metric from the quest's API endpoint with retry logic
-  const fetchUserMetric = useCallback(async (): Promise<number> => {
+  // Function to fetch user metric from the quest's API endpoint with retry logic.
+  // Returns both the raw metric (for progress display) and the server's
+  // authoritative `eligible` flag. The proof routes apply the per-verifier
+  // threshold server-side, so `eligible` — not a client-side `metric >= 1`
+  // heuristic — must decide whether a Claim button is shown. This is also the
+  // exact field `useClaimableQuestsCount` uses for the dashboard badge, so the
+  // badge and the quest cards can never disagree.
+  const fetchUserMetric = useCallback(async (): Promise<{
+    metric: number
+    eligible: boolean | null
+  }> => {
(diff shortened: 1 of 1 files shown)


pmoncada (approved):

pmoncada:
**Review: Approve ✅**

Solid hardening.
- Claim now gates on the server's authoritative `eligible` flag (falling back to `metric >= 1` only when the route does not report it) — same field the dashboard badge uses, so badge and cards cannot disagree.
- Stale-response guard (`cancelled` flag in the effect cleanup) correctly prevents a previous wallet's response from setting state after an account switch.
- Shared `refreshSingleQuestEligibility` fixes the "Link GitHub / Refresh discarded the result" staleness.
- No backend change; POST claim still re-verifies server-side.

PR #1375, opened 2026-06-10
fix(citizen): skip geocoder when location is empty to prevent Antarctica default

mmoncada1:
## Summary

Fixes citizens inadvertently appearing in Antarctica on the network map, plus related map data issues.

- **Empty location on save** — the edit modal called the geocoder even on a blank location field; Google returned zero results and the `-90,0 / "Antarctica"` fallback was written on-chain. Now we skip the geocoder when location is empty and store an empty string.
- **Legacy plain-text locations** — many citizens have their location stored as a plain place-name string (e.g. `"Austin, Texas"`) with no coordinates, because geocoding only ever ran in the edit modal, not at mint. These were all defaulting to the `-90,0` sentinel and stacking in Antarctica. The map service now geocodes these names server-side (cached) so they render in the correct place. `parseLocationData` was also made robust to all three stored formats (geocoded JSON object, JSON-encoded string, raw text) and now treats both `(0,0)` and `(-90,0)` as non-real sentinels.
- **Deleted / blank profiles** — the map query now excludes deleted (`view = ''`), unnamed (`name = ''`), and image-less (`image = ''`) citizens so they don't appear as pins.
- **Genuine no-location citizens** — the few citizens who truly have no location are scattered deterministically across the Antarctic landmass (lat -72 to -87) instead of stacking into one tall column at the pole.
- **Preview env gate** — `lib/map/index.ts` and `useMapData.ts` checked `NEXT_PUBLIC_ENV=prod`, which isn't set on Vercel previews, so the map always showed dummy data. Aligned to `NEXT_PUBLIC_CHAIN=mainnet` like the rest of the codebase.

## Test plan

- [ ] Edit a citizen profile, clear the location field, save → stored as empty string (not Antarctica)
- [ ] Edit a citizen profile with a valid location → geocoded correctly with lat/lng
- [ ] Open the map → legacy plain-text-location citizens appear in their real cities, not Antarctica
- [ ] Deleted, unnamed, and image-less citizens do not appear as pins
- [ ] Citizens with genuinely no location are spread across the Antarctic continent (not in the ocean, not stacked)

### DIFF
--- a/ui/components/subscription/CitizenMetadataModal.tsx
+++ b/ui/components/subscription/CitizenMetadataModal.tsx
@@ -317,25 +317,34 @@ export default function CitizenMetadataModal({ nft, selectedChain, setEnabled }:
 
               const cleanedCitizenData = cleanData(citizenData)
 
-              const locationDataRes = await fetch('/api/google/geocoder', {
-                method: 'POST',
-                headers: {
-                  'Content-Type': 'application/json',
-                },
-                body: JSON.stringify({
-                  location: cleanedCitizenData.location,
-                }),
-              })
-              const { data: locationData } = await locationDataRes.json()
-              const locationLat = locationData?.results?.[0]?.geometry?.location?.lat || -90
-              const locationLng = locationData?.results?.[0]?.geometry?.location?.lng || 0
-              const locationName = locationData?.results?.[0]?.formatted_address || 'Antarctica'
-              const citizenLocationData = {
-                lat: locationLat,
-                lng: locationLng,
-                name: locationName,
+              let cleanedLocationData: { lat: number; lng: number; name: string } | string = ''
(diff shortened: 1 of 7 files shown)


pmoncada (approved):
Check edits as well

pmoncada (approved):

pmoncada:
**Review: Approve ✅**

Good root-cause fix for the Antarctica default.
- Verified `cleanData` only escapes strings — it does **not** drop `lat:0/lng:0`, so the `{lat:0,lng:0}` object survives intact.
- Final write guard `cleanedLocationData ? JSON.stringify(...) : ''` works because the populated value is always a truthy object.
- Minor behavior note: unresolvable locations now land on **"null island" (0,0)** on the map instead of Antarctica — better than before and acknowledged in the PR, but still a visible pin.

pmoncada (approved):

PR #1378, opened 2026-06-11
feat(deprize): resolution, redemption & refund (M4)

pmoncada:
## Summary

Builds M4 (resolution, redemption & refund) on top of the merged M3 bet router — the close-out loop that is the **mainnet launch gate**: until now the market took money in with no way to pay winners or refund bettors. Full spec + runbook in [`docs/DEPRIZE_M4.md`](../blob/feature/deprize-m4-redeem/docs/DEPRIZE_M4.md).

- **`DePrizeRedeem`** (new, stateless, non-upgradeable, non-custodial): one-time `ctf.setApprovalForAll`, then `redeem(deprizeId)` pulls the caller's outcome tokens, redeems on the Gnosis CTF, and pays out **ETH**. The same code path serves winner payouts (`[0,…,1,…,0]`) and the cancellation/no-winner refund (`[1,1,…,1]` → 1/N per token, the disclosed parimutuel refund). `previewRedeem` mirrors the CTF's per-position floor division exactly. Payout is scoped to the redemption's WETH **balance delta**, so WETH parked in the helper can never leak into a claim (the M3 residual-sweep lesson, regression-tested). No registry-state gate: bettors can always bypass the helper and call the CTF directly, so the CTF payout vector is the single source of truth.
- **`DePrizeResolve.s.sol`** (new): the oracle is the MoonDAO multisig, baked irreversibly into the conditionId — `reportPayouts` must be submitted by the Safe itself, so a reporter *contract* cannot exist. This read-only script enforces everything a reporter would have: registry state → payout vector (`SETTLED`/`M1_RELEASED`/`M2_COMPLETE` → winner; `NO_WINNER`/`CANCELLED` → equal payout; `M2_FAILED` → refused, the CTF is already final), winner-index lookup, conditionId recomputation (catches wrong questionId/oracle/slot-count), already-reported guard — then prints the exact Safe transaction.
- **Interfaces extended**: `IConditionalTokens` (resolution/redemption/id-helper surface) and `ILMSRWithTWAP` (owner unwind surface: `pause`/`close`/`withdrawFees`/`transferOwnership`).
- **Migration 08**: now transfers LMSR ownership to the oracle multisig right after creation — the factory makes the deployer EOA the owner, but `close()`/`withdrawFees()` (the treasury-seed unwind) are `onlyOwner` and must be Safe-executable. Also prints a reminder to record the `questionId` (resolution needs it; it is not stored on-chain).
- **Deploy script** `DePrizeRedeem.s.sol` (no proxy — nothing to upgrade).

### Design decisions (recorded in `DEPRIZE_M4.md`)

- Refund = equal-payout report + the same redemption path; the design doc's combined one-click `refundAll` (CTF + JB cashOut) is dropped — the JB leg already exists via the M2 hook.
- A tradable "none fly by X date" outcome slot was considered and rejected (delays all payouts to delivery, converts the disclosed partial refund into a 100% loss for team bettors, touches the audited M3 router, oracle conflict).
- Unwound market proceeds (seed + fees) go to the treasury, **not** the JB project, to preserve the disclosed cancellation loss (double-count guard).

## Tests

`forge test --match-path 'test/deprize/*.t.sol'` → **121 passed** (30 new unit tests + 3 guarded fork tests).

- Unit suite runs against a `MockResolvingCTF` faithful to the deployed 0.5 source: msg.sender-derived conditionId, write-once payout vector, per-position floor division, burn-on-redeem, ERC-1155 approval checks + acceptance hooks. Covers winner/loser/refund paths, parimutuel skew, preview-equals-actual rounding, all guards (NotResolved/UnknownDePrize/NothingToRedeem/approval/RedeemFailed/reentrancy/double-redeem), stray-WETH regression, receiver-hook gating, and all 8 `buildReport` pre-flight cases (including that the emitted calldata actually resolves the condition when submitted by the oracle).
- Fork tests (skip unless `DEPRIZE_FORK_RPC` is set) run against the **real Arbitrum-Sepolia CTF** with a freshly-prepared condition and test oracle: winner redemption, cancellation 1/N redemption, and (with `DEPRIZE_FORK_FACTORY`) the full mainnet rehearsal — provision → bets via real `DePrizeMint` → lock → pause → settle → close → report → withdrawFees → treasury redeems inventory → bettors redeem — with ETH conservation asserted to the wei.

## Test plan

- [x] CI: unit suite green (fork tests no-op without `DEPRIZE_FORK_RPC`)
- [x] Run the guarded fork tests against Arbitrum Sepolia (incl. the full close-out loop with the LMSR factory)
- [ ] For the already-provisioned testnet market: deployer EOA `transferOwnership(multisig)`; verify the Safe's fallback handler accepts ERC-1155 before any `close()`
- [ ] Dry-run `DePrizeResolve.s.sol` against a testnet DePrize and submit the printed calldata from a test Safe

---

> [!NOTE]
> **High Risk**
> New on-chain redemption and oracle resolution paths handle user funds and irreversible CTF payouts; mistakes in runbook ordering (report before pause) or conditionId/questionId mismatch could strand value or enable arb against treasury inventory.
> 
> **Overview**
> **Closes the DePrize money-out loop (M4)** — the launch gate for paying winners and refunding bettors after M3 betting.
> 
> Adds **`DePrizeRedeem`**: a stateless helper that pulls a bettor’s CTF outcome tokens, redeems collateral, unwraps WETH, and pays **ETH**. One contract covers winner payouts and cancellation/no-winner refunds (equal `[1,1,…,1]` → 1/N per token). Payout uses a WETH balance **delta** so stray funds in the helper cannot inflate claims.
> 
> Replaces the planned **`DePrizeReporter`** with **`DePrizeResolve.s.sol`**: read-only pre-flight (registry state, conditionId vs `questionId`/oracle, already-reported, LMSR paused/closed) and Safe calldata for multisig `reportPayouts`.
> 
> **Provisioning & interfaces:** migration `08` transfers LMSR **ownership to the oracle** after create; CTF and LMSR interfaces gain resolution/redemption and owner unwind methods. Docs mark M4 complete and simplify refund vs combined `refundAll`.
> 
> **UI / testnet:** `deprize-play` gains resolution state, pause/close/withdraw runbook actions, oracle pre-flight resolve, and claim via `DePrizeRedeem` + config for Sepolia play market; small thirdweb/build fixes.
> 
> **Tests:** large `DePrizeRedeem.t.sol` unit suite + optional Arbitrum Sepolia fork close-out loop.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a9f111240644ebd724358d413216590004c5771. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/subscription-contracts/src/deprize/DePrizeRedeem.sol
+++ b/subscription-contracts/src/deprize/DePrizeRedeem.sol
@@ -45,6 +45,7 @@ contract DePrizeRedeem is ReentrancyGuard, IERC1155Receiver {
     error NothingToRedeem(uint256 deprizeId, address account);
     error RedeemFailed();
     error UnexpectedERC1155();
+    error SlotCountMismatch(uint256 deprizeId, uint256 teamIds, uint256 outcomeSlots);
 
     constructor(address registry_, address ctf_, address weth_) {
         if (registry_ == address(0) || ctf_ == address(0) || weth_ == address(0)) revert ZeroAddress();
@@ -136,7 +137,15 @@ contract DePrizeRedeem is ReentrancyGuard, IERC1155Receiver {
         if (dp.state == IDePrizeRegistry.DePrizeState.NONE) revert UnknownDePrize(deprizeId);
         conditionId = dp.ctfConditionId;
 
+        // The payout vector lives on the CTF and is indexed by the condition's
+        // outcome slots. Trusting `teamIds.length` alone would silently under-pay
+        // (skip real positions) or revert deep in `payoutNumerators` if the
+        // registry ever diverges from the on-chain condition, so assert they
+        // agree up front and fail loudly on inconsistent provisioning.
         uint256 n = dp.teamIds.length;
+        uint256 slots = ctf.getOutcomeSlotCount(conditionId);
+        if (n != slots) revert SlotCountMismatch(deprizeId, n, slots);
+
         positionIds = new uint256[](n);
         for (uint256 i = 0; i < n; i++) {
             positionIds[i] = ctf.getPositionId(address(weth), ctf.getCollectionId(bytes32(0), conditionId, 1 << i));
--- a/subscription-contracts/src/deprize/DePrizeRedeem.sol
+++ b/subscription-contracts/src/deprize/DePrizeRedeem.sol
@@ -1,0 +1,171 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import {IERC1155Receiver} from "@openzeppelin/contracts/token/ERC1155/IERC1155Receiver.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+
+import {IDePrizeRegistry} from "./IDePrizeRegistry.sol";
+import {IConditionalTokens} from "./interfaces/IConditionalTokens.sol";
+import {IWETH} from "./interfaces/IWETH.sol";
+
+/// @title DePrizeRedeem
+/// @notice Bettor redemption helper for resolved DePrize markets. A single
+///         `redeem` call pulls the caller's outcome tokens (one-time
+///         `ctf.setApprovalForAll(address(this), true)` required), redeems them
+///         on the Gnosis CTF, unwraps the WETH payout, and pays the caller ETH.
+///
+///         Serves BOTH payout paths with the same code:
+///         - winner resolution (`[0,…,1,…,0]`): winning tokens pay full value;
+///         - no-winner / cancellation resolution (`[1,1,…,1]`): every token pays
+///           1/N (the disclosed parimutuel refund).
+///
+/// @dev Deliberately thin, stateless between calls, and NON-custodial:
+///      redemption is already permissionless on the CTF (bettors hold the
+///      ERC-1155 outcome tokens directly since M3 and can always call
+///      `ctf.redeemPositions` themselves — they would just receive WETH instead
+///      of ETH). Because anyone can bypass this contract, it adds no lifecycle
+///      gates: the CTF payout vector (`payoutDenominator`) is the sole source
+///      of truth for when/what redemption pays. It is also intentionally
+///      non-upgradeable — no state worth migrating, smaller trust surface.
+contract DePrizeRedeem is ReentrancyGuard, IERC1155Receiver {
+    IDePrizeRegistry public immutable registry;
+    IConditionalTokens public immutable ctf;
+    IWETH public immutable weth;
+
+    /// @dev True only while `redeem` is pulling the caller's outcome tokens;
+    ///      gates the ERC-1155 acceptance hooks (no unsolicited deposits).
+    bool private _inRedeem;
+
+    event Redeemed(uint256 indexed deprizeId, address indexed account, uint256 payout);
+
+    error ZeroAddress();
+    error UnknownDePrize(uint256 deprizeId);
+    error NotResolved(uint256 deprizeId);
+    error NothingToRedeem(uint256 deprizeId, address account);
+    error RedeemFailed();
+    error UnexpectedERC1155();
+
+    constructor(address registry_, address ctf_, address weth_) {
+        if (registry_ == address(0) || ctf_ == address(0) || weth_ == address(0)) revert ZeroAddress();
+        registry = IDePrizeRegistry(registry_);
+        ctf = IConditionalTokens(ctf_);
+        weth = IWETH(weth_);
+    }
+
+    // ---------------------------------------------------------------------
+    // Redemption
+    // ---------------------------------------------------------------------
+
+    /// @notice Redeem all of `msg.sender`'s outcome tokens for `deprizeId` and
+    ///         pay out the proceeds in ETH. Requires the caller to have approved
+    ///         this contract on the CTF (`setApprovalForAll`). Reverts until the
+    ///         oracle has resolved the condition.
+    function redeem(uint256 deprizeId) external nonReentrant {
+        (bytes32 conditionId, uint256[] memory positionIds) = _positions(deprizeId);
+        if (ctf.payoutDenominator(conditionId) == 0) revert NotResolved(deprizeId);
+
+        // Find the caller's nonzero positions.
+        uint256 n = positionIds.length;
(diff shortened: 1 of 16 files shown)


mmoncada1 (changes requested):
fix failing test

subscription-contracts/src/deprize/DePrizeRedeem.sol
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /// @dev Resolve a DePrize to its condition id and per-outcome-slot position
+    ///      ids (slot i = `registry.teamIds(deprizeId)[i]`, the M3 convention).
+    function _positions(uint256 deprizeId) private view returns (bytes32 conditionId, uint256[] memory positionIds) {
+        IDePrizeRegistry.DePrize memory dp = registry.getDePrize(deprizeId);
+        if (dp.state == IDePrizeRegistry.DePrizeState.NONE) revert UnknownDePrize(deprizeId);
+        conditionId = dp.ctfConditionId;
+
+        uint256 n = dp.teamIds.length;
```
pmoncada, line 145: **Helper trusts `registry.teamIds.length` for slot count (important).** `n` is derived from the registry, but the payout vector lives on the CTF. If `teamIds.length` ever diverges from the condition's actual `outcomeSlotCount`, `redeem`/`previewRedeem` either silently skip real positions (under-pay) or revert via out-of-bounds `payoutNumerators`. The PR adds `getOutcomeSlotCount` to the interface but never uses it — suggest asserting `n == ctf.getOutcomeSlotCount(conditionId)` in `_positions` so inconsistent provisioning fails loudly.

ui/pages/deprize-play.tsx
```diff
@@ -989,58 +1126,179 @@ export default function DePrizePlay() {
...
+      })
+      if (
+        conditionId &&
+        computed.toLowerCase() !== conditionId.toLowerCase()
+      ) {
+        throw new Error(
+          `Pre-flight: conditionId mismatch — keccak(yourAddress, questionId, ${MAX_OUTCOMES}) != the market's condition. Check the question id and that you are the oracle.`
+        )
+      }
+      if (payoutDen && payoutDen > 0n) {
```
pmoncada: **`resolve()` pre-flight guards read possibly-stale React state (important).** `payoutDen` and `stage` come from periodic-refresh component state, so both the 'already resolved' and 'pause before report' guards can be bypassed by a race (acting on a stale snapshot). Reporting against a live market is the exact 'free trades against the known outcome' hazard the script hard-aborts on. The `getConditionId` check just above is a fresh `rpcRead` — these two should be too (re-read `payoutDenominator` and `stage()` immediately before sending).

ui/pages/deprize-play.tsx
```diff
@@ -989,58 +1126,179 @@ export default function DePrizePlay() {
...
+  const resolved = payoutDen !== undefined && payoutDen > 0n
+  const winningIndex = resolved
+    ? payoutNums.findIndex((n, _, arr) => n > 0n && arr.filter((x) => x > 0n).length === 1)
+    : -1
+  const isRefundVector =
+    resolved && payoutNums.length > 0 && payoutNums.every((n) => n > 0n)
+  const claimable = resolved
+    ? outcomes.reduce((acc, o, i) => {
+        if (!Number.isFinite(o.balance) || o.balance <= 0) return acc
+        const balWei = BigInt(Math.floor(o.balance * Number(UNIT)))
```
pmoncada: **`claimable` estimate loses precision via float→BigInt (nit).** `o.balance` is a JS float, so `Math.floor(o.balance * 1e18)` and the trailing `Number(payout)/1e18` both lose precision for large positions. Display-only (the authoritative `previewRedeem` is exact), but consider sourcing the balance in wei (BigInt) so the headline number can't disagree with what `redeem` actually pays.

ui/pages/deprize-play.tsx
```diff
@@ -989,58 +1126,179 @@ export default function DePrizePlay() {
...
 
   const isClosed = stage === MarketStage.Closed
+  // CTF-level resolution is what actually gates redemption (market stage is
+  // only the trading state). claimable = what the connected wallet's positions
+  // redeem for under the reported payout vector (mirrors previewRedeem).
+  const resolved = payoutDen !== undefined && payoutDen > 0n
+  const winningIndex = resolved
+    ? payoutNums.findIndex((n, _, arr) => n > 0n && arr.filter((x) => x > 0n).length === 1)
+    : -1
+  const isRefundVector =
```
pmoncada: **`isRefundVector` assumes exactly `MAX_OUTCOMES` live slots (nit).** `payoutNums` is padded to `MAX_OUTCOMES` (3) with non-existent/errored slots as `0n`. For a market with fewer than 3 real outcomes, a genuine `[1,1,...]` refund would have a trailing `0n` and be misclassified as 'not a refund' (and `winningIndex` would misfire). Fine under the fixed 3-outcome convention, but brittle if the slot count ever changes.

ui/lib/thirdweb/client.ts
```diff
@@ -27,8 +27,15 @@ const client = createThirdwebClient({
   },
 })
 
+// Prefer the secret key (higher limits, server-only endpoints) but fall back
```
pmoncada, line 30: **`serverClient` secret env var is `NEXT_PUBLIC_`-prefixed (nit).** Two issues: (a) when the secret is absent, `serverClient` silently becomes a public-clientId client with lower limits/permissions and no signal; (b) any `NEXT_PUBLIC_*` value is inlined into the browser bundle — a real secret key here would be exposed client-side. Pre-existing, but worth fixing the naming (drop `NEXT_PUBLIC_`) while touching this file.

pmoncada (commented):
**Review summary — deprize M4 resolution/redemption/refund**

High-quality, well-tested PR. The core money-out contract is correct on the dimensions that matter: `nonReentrant`, payout scoped to the WETH balance *delta*, receiver gated to mid-redeem CTF pulls, burn-then-pay so double-redeem reverts, and `previewRedeem` mirrors the CTF's per-position floor division. Resolution correctly relies on the CTF's write-once `payoutDenominator`. **No blocking financial-correctness bug found.** The two important items are defensive hardening: assert the slot count against `getOutcomeSlotCount` rather than trusting `registry.teamIds.length`, and make the `resolve()` pre-flight guards re-read on-chain state instead of stale React state. See inline comments.

mmoncada1 (approved):

PR #1380, opened 2026-06-12
feat(landing): add new landing components and animations

mmoncada1:
- Introduced several new components for the landing page, including CountUp, CtaButton, FinalCTA, GovernanceSection, JourneyTimeline, LandingHero, LaunchpadShowcase, MissionStory, and PartnersMarquee.
- Implemented animations using framer-motion for enhanced user experience.
- Updated package.json to include framer-motion as a dependency and added resolutions for walletconnect.
- Enhanced tailwind.config.js with new animations and styles for the landing components.


ui/components/home/landing/PillarsSection.tsx
```diff
@@ -0,0 +1,186 @@
...
+    setGlow((g) => ({ ...g, visible: false }))
+  }
+
+  return (
+    <Link
+      ref={ref}
+      href={pillar.link}
+      onMouseMove={handleMouseMove}
+      onMouseLeave={reset}
+      style={{ transform }}
```
pmoncada: **Per-`mousemove` `setState` causes heavy re-renders (important).** Each pointer move triggers two `setState` calls, re-rendering the whole `TiltCard` (and its `<Image>`/SVG children) on every event — up to 6 cards at once. Prefer writing the transform/glow directly to the DOM node via the existing `ref` (`el.style.transform = ...`, a CSS var for the glow) so the tilt updates without React re-renders. Also gate the tilt behind `useReducedMotion()`.

ui/tailwind.config.js
```diff
@@ -23,6 +23,9 @@ module.exports = {
         fadeInSlowTo75: 'fadeInTo75 1s ease-in',
         notification: 'notification 5s',
         highlight: 'highlight 3s ease-out',
+        marquee: 'marquee 45s linear infinite',
```
pmoncada, line 26: **New CSS animations ignore `prefers-reduced-motion` (important).** Unlike the framer-motion pieces, `marquee`/`marquee-reverse`/`float` run continuously regardless of the user's reduced-motion setting (the marquee only pauses on hover). Add `motion-reduce:animate-none` to the elements using them, or wrap the keyframes in a `@media (prefers-reduced-motion: no-preference)` guard.

ui/components/home/landing/LandingHero.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <div className="absolute inset-0 bg-[radial-gradient(ellipse_at_50%_-20%,rgba(66,94,235,0.28),transparent_60%)]" />
+        <Starfield className="opacity-80" />
+
+        {/* Content */}
+        <motion.div
+          style={{ opacity: contentOpacity, y: contentY }}
+          className="relative z-10 mx-auto flex w-full max-w-[1400px] flex-1 flex-col items-center justify-center px-5 py-24 text-center md:px-10"
+        >
+          <motion.div variants={container} initial="hidden" animate="visible">
+            <motion.h1 variants={item} className="font-GoodTimes leading-none text-white">
```
pmoncada, line 79: **Hero headline renders invisible until JS executes (important — LCP/SEO/no-JS).** `LandingHero` is statically imported (SSR'd) but the `<h1>` starts at `initial="hidden"` (`opacity:0`), so the main heading is invisible in the server HTML and only appears after hydration. Consider animating only `y`/transform (keep `opacity:1`), or render visible and animate in via a mount flag.

ui/components/home/landing/GovernanceSection.tsx
```diff
@@ -0,0 +1,147 @@
...
+                </span>
+              </>
+            }
+            description="Everything at MoonDAO is proposed, governed, and created by its members. Lock $MOONEY to become a voter and co-govern the treasury — no gatekeepers, no permission needed."
+          />
+
+          <Reveal delay={0.25}>
+            <button
+              onClick={() => {
+                navigator.clipboard.writeText(MOONEY_ADDRESS)
```
pmoncada: **Clipboard write has no guard or error handling (important).** `navigator.clipboard` is `undefined` in insecure/older contexts and `writeText` returns a promise that can reject — here it would throw/reject unhandled while still showing a success toast. Wrap in `try/await` (or `.then/.catch`) and only toast success on resolve.

ui/components/home/landing/LandingHero.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+export default function LandingHero() {
+  const sectionRef = useRef<HTMLElement>(null)
+  const reduceMotion = useReducedMotion()
+
+  const { scrollYProgress } = useScroll({
+    target: sectionRef,
+    offset: ['start start', 'end start'],
+  })
+  const bgY = useTransform(scrollYProgress, [0, 1], ['0%', reduceMotion ? '0%' : '22%'])
```
pmoncada: **Parallax translate exceeds the background's scale headroom (important).** `scale-[1.12]` gives only ~6% overflow per edge, but `bgY` translates the background down ~22% of its height on scroll, which can drag the top edge into view and expose a gap (partly masked by the top gradient). Either reduce the travel (~6%) or increase the scale so travel ≤ available headroom.

ui/components/home/LaunchpadSection.tsx
```diff
@@ -100,7 +100,7 @@ export default function LaunchpadSection() {
                   <div>
                     <h4 className="text-white font-semibold text-lg mb-1">100% Transparent</h4>
                     <p className="text-white/70 text-sm md:text-base">
-                      All transactions are onchain, ensuring everyone can see how funds are spent.
+                      Transactions are onchain, so everyone can see how funds are spent.
```
pmoncada: **This copy edit is orphaned by this same PR (nit).** The PR removes the only import of `LaunchpadSection` from `pages/index.tsx`, and it isn't referenced anywhere else, so this change has no effect. Either drop it or delete the now-unused component (and the other orphaned `home/` components: `Hero`, `Callout1-3`, `Feature`, `Timeline`, `PartnerSection`, `PreFooter`).

ui/components/home/landing/CountUp.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const inView = useInView(ref, { once: true, margin: '-40px' })
+
+  useEffect(() => {
+    if (!inView || !ref.current) return
+    const controls = animate(0, to, {
+      duration,
+      ease: [0.16, 1, 0.3, 1],
+      onUpdate: (value) => {
+        if (ref.current) {
+          ref.current.textContent = `${prefix}${new Intl.NumberFormat(
```
pmoncada: **`CountUp` mutates `textContent` directly (nit).** Writing to `textContent` of a node whose children React also controls (`{prefix}0{suffix}`) works here only because the stats are static; any parent re-render would snap the number back to `0`. Consider driving the value through a framer-motion `useMotionValue`/`<motion.span>` or React state to stay declarative.

ui/components/home/landing/Starfield.tsx
```diff
@@ -0,0 +1,115 @@
...
+          if (s.y > height + 2) {
+            s.y = -2
+            s.x = Math.random() * width
+          }
+        }
+      }
+      ctx.globalAlpha = 1
+    }
+
+    const loop = (t: number) => {
```
pmoncada, line 84: **`Starfield` rAF runs continuously even off-screen (nit).** The `requestAnimationFrame` loop never pauses when scrolled out of view, and there are two `Starfield` instances (hero + final CTA), so two loops run all session. Cleanup on unmount is correct, but consider pausing via an `IntersectionObserver`. Also `prefersReducedMotion` is captured once at mount and won't react to a live preference change.

pmoncada:
fix failing test in another PR

pmoncada (approved):

PR #1382, opened 2026-06-12
Allow EU/EEA/UK visitors to browse while blocking on-chain profile creation

mmoncada1:
## Summary
- Replaces the full EU/UK site block with a GDPR-compliant workaround: visitors from EU, EEA, and UK regions can browse the site but cannot start citizen or team creation flows that permanently store personal data on chain.
- Adds geo restriction helpers (`isEUCountry`, `useRegionRestriction`) and extends `/api/geo/country` with a `restricted` flag.
- Shows a dedicated `RegionRestrictedNotice` on `/citizen` and `/team` for restricted regions instead of the creation wizards.

## Test plan
- [ ] Confirm Vercel/edge firewall no longer fully blocks EU/EEA/UK traffic (required for this change to take effect)
- [ ] Visit `/citizen` from an EU/EEA/UK IP (or mock `/api/geo/country` to return `restricted: true`) and verify the restriction notice appears
- [ ] Visit `/team` from a restricted region and verify the restriction notice appears
- [ ] Visit `/citizen` and `/team` from a non-restricted region and verify normal creation flows still work
- [ ] Verify `/join`, citizen/team profile pages, and other browse-only pages remain accessible from restricted regions
- [ ] Verify `/api/geo/country` still returns `country` for the cookie banner and now includes `restricted`
- [ ] Verify UK (`GB`) is treated the same as EU/EEA for restriction


ui/pages/citizen/index.tsx
```diff
@@ -39,12 +45,18 @@ export default function Join() {
...
       />
-      <CreateCitizen
-        selectedChain={selectedChain}
-        setSelectedTier={handleExitFlow}
-        freeMintProp={freeMint}
-        inviteToken={inviteToken}
-      />
+      {/* Wait until geo resolves so we never flash the creation flow to a
+          restricted visitor. */}
+      {isResolvingRegion ? null : isRestricted ? (
```
pmoncada: **Enforcement is client-side only (important).** This React conditional (mirrored in `team/index.tsx`) is the only thing gating restricted users. The actual personal-data writes — `mintTo` via `prepareContractCall` and the supporting API routes (`/api/mission/freeMint`, `/api/image-gen/citizen-image`, `/api/typeform/response`, IPFS pin) — perform no geo check, so the gate is trivially bypassed (disabled JS, calling endpoints directly, or when the geo fetch fails). For a GDPR control, add defense-in-depth: enforce `getCountryFromHeaders(req)` + `isEUCountry()` server-side in the data-persistence routes and reject with 451.

ui/lib/geo/useRegionRestriction.ts
```diff
@@ -0,0 +1,46 @@
...
+    {
+      revalidateOnFocus: false,
+      dedupingInterval: 1000 * 60 * 60, // geo is stable per session
+      shouldRetryOnError: false,
+    }
+  )
+
+  return {
+    country: data?.country ?? null,
+    isRestricted: Boolean(data?.restricted),
```
pmoncada, line 52: **Geo hook fails open (important).** With `shouldRetryOnError: false`, if `/api/geo/country` errors, `isLoading` becomes `false` while `data` stays `undefined`, so `isRestricted === false` and the creation flow renders to a potentially-EU user. Missing/unknown geo headers do the same. Failing open is fine for the cookie banner, but for a hard compliance block consider failing closed for the creation path (treat unresolved/errored geo as restricted) and/or allowing a retry.

ui/lib/geo/useRegionRestriction.ts
```diff
@@ -0,0 +1,46 @@
...
+  // True while geo is still being resolved. Gate destructive UI on this so we
+  // don't briefly flash a creation flow before discovering the user is in a
+  // restricted region.
+  isLoading: boolean
+}
+
+/**
+ * Resolves whether the current visitor is in a GDPR-restricted (EU/EEA) region.
+ *
+ * The previous behavior fully blocked these visitors at the edge. We now let
```
pmoncada, line 28: **'Previously blocked at the edge' doesn't exist in this repo (please confirm).** There's a leftover `pages/451.tsx` and a `Layout.tsx` special-case for `/451`, but no `middleware.ts` or in-code redirect to `/451` anywhere in `ui/`. If the old edge block lives in a Vercel WAF / firewall rule that's still active, EU/EEA/UK visitors are hard-blocked before they ever reach `/citizen` or `/team`, making this PR a no-op. Please confirm the external rule is removed in lockstep, and consider deleting the now-dead `451.tsx` / Layout case.

ui/pages/team/index.tsx
```diff
@@ -15,9 +17,23 @@ export default function TeamJoin() {
...
+  // EU/EEA visitors may browse the site but cannot permanently store personal
+  // data on chain (GDPR), so they don't get the team creation flow.
+  const { isRestricted, isLoading: isResolvingRegion } = useRegionRestriction()
+
   // Ensure default chain settings
   useChainDefault()
 
+  // Wait until geo resolves so we never flash the creation flow to a
+  // restricted visitor.
+  if (isResolvingRegion) {
```
pmoncada, line 46: **Returns `null` during geo resolution (nit).** Unlike `citizen/index.tsx` (which keeps `<Head>` mounted and only swaps the body), the team page returns `null` while waiting on the client geo fetch — losing `<Head>`/SEO metadata and flashing blank/CLS on every visit. Prefer rendering `Head` + a lightweight skeleton during `isResolvingRegion` for consistency.

ui/lib/geo/index.ts
```diff
@@ -41,6 +41,52 @@ export function getCountryFromHeaders(req: NextApiRequest): string | null {
...
+  'IS', // Iceland (EEA)
+  'LI', // Liechtenstein (EEA)
+  'NO', // Norway (EEA)
+  'GB', // United Kingdom (UK-GDPR)
+])
+
+// Returns true when the country code belongs to a GDPR-restricted region
+// (EU / EEA / UK). Unknown/null countries return false so we never block on
+// missing geo data.
+export function isEUCountry(countryCode: string | null | undefined): boolean {
```
pmoncada, line 89: **EU/EEA set misses UK crown dependencies / Gibraltar (nit).** Vercel/Cloudflare emit distinct ISO codes for UK-GDPR-adjacent jurisdictions that won't match `GB`: Gibraltar (`GI`), Jersey (`JE`), Guernsey (`GG`), Isle of Man (`IM`). If the intent includes UK-GDPR coverage, consider adding these. (Switzerland `CH` has its own FADP — reasonable to leave out, but worth a conscious decision.)

pmoncada (commented):
**Review summary — EU/EEA/UK browse gating**

Clean, well-typed UI implementation: `isEUCountry`/`EU_EEA_COUNTRIES` is correct, the SWR hook is reasonable, and page-level gating covers the only two import sites of `CreateCitizen`/`CreateTeam`. The substantive concern is that the whole compliance guarantee rests on a **client-side, fail-open** check with **no server-side enforcement** on the data-writing paths — for a GDPR control it's a UX nudge rather than a block. Also please confirm the "previously blocked at the edge" rule (Vercel WAF?) is being removed in lockstep, since no such block exists in the repo. See inline comments for specifics (2 important, 3 nits).

pmoncada (commented):
**Review summary — EU/EEA/UK browse gating**

Clean, well-typed UI implementation: `isEUCountry`/`EU_EEA_COUNTRIES` is correct, the SWR hook is reasonable, and page-level gating covers the only two import sites of `CreateCitizen`/`CreateTeam`. The substantive concern is that the whole compliance guarantee rests on a **client-side, fail-open** check with **no server-side enforcement** on the data-writing paths — for a GDPR control it's a UX nudge rather than a block. Also please confirm the "previously blocked at the edge" rule (Vercel WAF?) is being removed in lockstep, since no such block exists in the repo. See inline comments for specifics (2 important, 3 nits).

pmoncada (approved):

pmoncada:
test in prod w/ a VPN in prod

PR #1386, opened 2026-06-15
feat(team): show stats, jobs preview, and marketplace to all visitors

mmoncada1:
## Summary

- **Stats bar** (members / open jobs / marketplace items / active missions) is now visible to all visitors — no citizenship required
- **Marketplace listings** are fully visible to everyone (removed `canViewContent` gate)
- **Jobs board** is now visible to everyone; non-citizens see a blurred preview with job count and a "Become a Citizen" CTA (existing `TeamJobs` preview logic was already in place)
- **TeamMembers, TeamTreasury, TeamMissions** remain citizen-only — no change

## Test plan

- [ ] Visit a team page as a logged-out/non-citizen user — stats, jobs preview, and marketplace should all be visible
- [ ] Verify jobs show the blur/lock preview for non-citizens, with full details for citizens/managers
- [ ] Verify TeamMembers, TeamTreasury, TeamMissions are still hidden for non-citizens
- [ ] Verify managers/owners still see all sections and edit controls as before

Made with [Cursor](https://cursor.com)

PR #1387, opened 2026-06-15
feat(marketplace): restore share/copy-link button on listings

mmoncada1:
## Summary

Restores the share button that was previously removed from marketplace listing cards.

- Each listing card now has a `🔗` (link) icon button in the footer, visible to all users
- Clicking it copies a deep-link URL to the clipboard: `/team/<slug>?listing=<id>` (e.g. `/team/moondao?listing=31`)
- The team slug is derived from the team name if available, falling back to the numeric team ID
- The existing `?listing=` query param handling already auto-opens the purchase modal and scrolls to the correct card on page load — no server-side changes needed
- The visibility bug that would have blocked these links for non-citizens is fixed in PR #1386

## Test plan

- [ ] Visit a team page as any user — each listing card should show a link icon in the bottom-right
- [ ] Click the icon — toast should confirm "Link copied to clipboard"
- [ ] Paste and navigate to the copied URL — the listing's buy modal should open automatically
- [ ] Confirm managers still see pencil + trash icons alongside the share icon

### DIFF
--- a/ui/components/subscription/TeamListing.tsx
+++ b/ui/components/subscription/TeamListing.tsx
@@ -223,14 +223,22 @@ export default function TeamListing({
             <div className="flex items-center gap-2">
               <button
                 id="share-listing-button"
-                onClick={(e) => {
+                onClick={async (e) => {
                   e.stopPropagation()
                   const origin = typeof window !== 'undefined' ? window.location.origin : 'https://www.moondao.com'
                   const teamSlug = teamNFT?.metadata?.name
                     ? generatePrettyLink(teamNFT.metadata.name)
                     : String(listing.teamId)
-                  navigator.clipboard.writeText(`${origin}/team/${teamSlug}?listing=${listing.id}`)
-                  toast.success('Link copied to clipboard.')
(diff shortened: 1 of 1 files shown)


pmoncada (commented):
**Review (automated pass by pmoncada)** — straightforward feature, two things to address.

**Heads-up: branch appears based on stale `main`**
- The diff's context for `TeamListing.tsx` shows `import { useWallets } from '@privy-io/react-auth'` and `useContext` in the import block, but neither exists in the current `main` version of this file. That means the branch diverged from `main` and a rebase will likely be needed — please rebase and confirm the merge doesn't reintroduce now-removed imports/usage.

**`navigator.clipboard` handling**
- `navigator.clipboard.writeText(...)` isn't awaited and has no error handling, yet the success toast fires unconditionally. In a non-secure context or if the write rejects, the user sees "Link copied" but nothing was copied. Consider awaiting and only toasting on success (with a fallback/error toast).

**Good**
- SSR-guarded `window.location.origin` with the `moondao.com` fallback is correct.
- Share button visible to all, edit/trash still gated behind `editable` — the JSX restructure (fragment for the editable group) is balanced.

pmoncada (approved):

PR #1388, opened 2026-06-15
fix(team): manage team UI and restore Safe signer data

mmoncada1:
## Summary
- Rename "Manage Members" to "Manage Team" on the button and modal to avoid confusion when adding managers
- Remove duplicate close button from the manage team modal (shared `Modal` already provides one)
- Restore team Safe signer addresses and pending transactions by falling back to client-side `useSafe` owners when SSR `safeOwners` is empty

## Test plan
- [ ] Open a team page as a manager and confirm the button/modal say "Manage Team"
- [ ] Open the manage team modal and confirm only one close (X) button appears
- [ ] On a team with a Safe multisig, confirm signer addresses appear in the Treasury section
- [ ] Connect a Privy wallet that is a Safe signer and confirm pending transactions are visible and can be signed

### DIFF
--- a/ui/components/subscription/TeamManageMembers.tsx
+++ b/ui/components/subscription/TeamManageMembers.tsx
@@ -1,5 +1,4 @@
 import { hatIdDecimalToHex } from '@hatsprotocol/sdk-v1-core'
-import { XMarkIcon } from '@heroicons/react/24/outline'
 import { TrashIcon } from '@heroicons/react/24/outline'
 import { useWallets } from '@privy-io/react-auth'
 import { DEFAULT_CHAIN_V5, HATS_ADDRESS } from 'const/config'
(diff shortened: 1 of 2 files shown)


pmoncada (commented):
**Review (automated pass by pmoncada)** — clean, low-risk, and fixes a real bug. 👍

**Good**
- The previous `safeOwners.includes(address || '')` was case-sensitive against checksummed addresses, so the new case-insensitive `.some()` comparison is a genuine bug fix for the signer-gated UI.
- Falling back to client-side `useSafe` owners when the SSR `safeOwners` comes back empty is a sensible resilience fix, and `resolvedSafeOwners` is threaded through consistently to both `TeamTreasury` instances.
- Removing the duplicate close button (the shared `Modal` already provides one) and the rename are straightforward.

**Note**
- This file (`ui/pages/team/[tokenIdOrName].tsx`) is also edited by PR #1386 in an overlapping region — whichever merges first will require the other to rebase. Worth sequencing them.

pmoncada (approved):

PR #1389, opened 2026-06-15
Fix citizen profile listings spacing and governance stats

mmoncada1:
## Summary
- Fix Newest Listings card spacing on the citizen profile by removing the oversized wrapper width so cards sit next to each other with consistent gaps.
- Add tooltips to the Governance section explaining $MOONEY, Voting Power, and Votes.
- Fix the Votes count mismatch by using the same on-chain + Snapshot vote logic as the Votes XP quest via a shared helper and public `/api/xp/votes-count` endpoint.

## Test plan
- [ ] Open a citizen profile and confirm Newest Listings cards are adjacent without large gaps between them.
- [ ] Hover/click the `?` tooltips in the Governance section and verify copy for $MOONEY, Voting Power, and Votes.
- [ ] Compare the profile Votes number with the Votes quest metric for the same wallet and confirm they match.
- [ ] Verify `/api/xp/votes-count?address=<wallet>` returns the expected count for wallets with on-chain and/or Snapshot votes.

### DIFF
--- a/ui/lib/xp/votes.ts
+++ b/ui/lib/xp/votes.ts
@@ -1,0 +1,83 @@
+import { gql, request as gqlRequest } from 'graphql-request'
+import { Address } from 'thirdweb'
+import {
+  DEFAULT_CHAIN_V5,
+  NON_PROJECT_PROPOSAL_TABLE_NAMES,
+  PROPOSALS_TABLE_NAMES,
+} from 'const/config'
+import queryTable from '@/lib/tableland/queryTable'
+import { getChainSlug } from '@/lib/thirdweb/chain'
+
+// Server-side only. Centralizes the canonical vote count used for both the
+// "Votes" XP quest and the Citizen profile so the two never drift apart.
+
+const chain = DEFAULT_CHAIN_V5
+const chainSlug = getChainSlug(chain)
+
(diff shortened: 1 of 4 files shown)


pmoncada (commented):
**Review (automated pass by pmoncada)** — looks solid overall, a few minor notes.

**Good**
- Nice refactor extracting the duplicated vote-counting logic into a shared `ui/lib/xp/votes.ts` so the XP quest and the profile can't drift. This directly addresses the count mismatch.
- The new public `/api/xp/votes-count` endpoint validates the address with `ethersUtils.isAddress` before it's interpolated into the Tableland SQL, so the raw string interpolation stays injection-safe (hex only). The cache header is a nice touch.

**Minor / non-blocking**
- Skeleton width changed to `w-56` (224px) but the loaded card wrapper now has no fixed width (`flex-shrink-0` only). Please confirm the real `TeamListing` cards render at ~224px so there isn't a visible width/layout shift when skeletons swap to content.
- The votes fetch uses a manual `fetch` + `useState`/`useEffect`. Per repo conventions (SWR / react-query for client data) consider `useSWR` here for caching + dedupe — not blocking.
- `{votesCount ?? 0}` renders `0` during loading (previously `votes?.length` was blank). Minor UX; a skeleton or `—` while `undefined` would read better.

pmoncada (approved):

PR #1391, opened 2026-06-16
Coinbase onramp

mmoncada1:
## Summary

Replaces the legacy hosted Coinbase widget with a unified, region-aware fiat onramp so users can fund their wallet with **Apple Pay (Coinbase Headless Onramp v2)** in the US and **MoonPay** elsewhere, ahead of the June 30 deprecation of the old Coinbase flow. Adds a shared `FundOnramp` / `FundOnrampModal` provider picker and wires it consistently into every funding entry point.

Also upgrades **Privy to v3** (now using Privy's in-app fiat onramp for MoonPay) and includes the build/CI, CSP, and component-test fixes that upgrade required.

### Onramp
- **Coinbase Headless Onramp (v2)** — embedded Apple Pay sheet for US users (`CBHeadlessOnramp`), with USD amount input, live ETH quoting, and a post-purchase confirmation screen.
- **MoonPay** — in-app fiat onramp (`MoonPayOnramp`) for non-US users and as a fallback when Apple Pay isn't supported.
- **Unified picker** (`FundOnramp` / `FundOnrampModal`) — defaults to Coinbase for US, MoonPay abroad, shows the recommended provider first, and detects region via `/api/coinbase/region`.
- **Return / auto-resume** — `OnrampReturnHandler` + JWT-scoped `useOnrampAutoTransaction` bring users back to their flow after a redirect (non-US Coinbase / MoonPay) or in-page reload (US Apple Pay) and safely resume the original mint/contribution.

### Entry points migrated to the new onramp
- Global header wallet **Fund** button (`PrivyConnectWallet`) — with free-form USD amount input.
- Dashboard **`WalletInfoCard`** Fund button.
- **`CreateCitizen`** and **`CreateTeam`** checkout (form-cache + auto-resume so progress survives the onramp round-trip).
- **`MissionContributeModal`** contribution flow.

### Privy v3 upgrade
- Upgrade `@privy-io/react-auth` to v3 and route MoonPay through Privy's in-app fiat onramp.
- Build/CI fixes for v3's optional transitive deps (alias unused Solana/stablelib/connectors, `ignore-engines`, clean-install cache fixes).
- Component-test fixes so Privy v3's remote-config fetch no longer crashes mounted Cypress specs.

### Infra
- New API routes: `/api/coinbase/create-onramp-order`, `/api/coinbase/region`.
- CSP / iframe `connect-src` + `frame-src` updates for the Coinbase risk SDK and Apple Pay.
- Apple Pay domain-association file hosted at `/.well-known/` for `www.moondao.com`.

## Test plan
- [x] US: Fund via Apple Pay from header button, dashboard, Create Citizen, Create Team, and a mission contribution — confirm balance updates and the original action resumes.
- [x] Non-US (or VPN): confirm MoonPay is the default and the redirect returns to the originating flow.
- [x] Confirm the recommended provider renders first and copy is accurate per region.
- [x] Verify Apple Pay domain verification works on production (`www.moondao.com`).
- [x] CI: Cypress component tests (mission specs) pass under Privy v3.


ui/components/privy/PrivyConnectWallet.tsx
```diff
@@ -989,6 +802,17 @@ export function PrivyConnectWallet({ citizenContract, type }: PrivyConnectWallet
                   />
                 )}
 
+                {address && (
+                  <FundOnrampModal
+                    enabled={fundModalOpen}
+                    setEnabled={setFundModalOpen}
+                    address={address}
+                    selectedChain={selectedChain}
+                    ethAmount={0}
```
pmoncada, line 811: **High — broken funding path.** This opens the combined onramp with `ethAmount={0}` and no `allowAmountInput`. For US users it defaults to `CBHeadlessOnramp`, which has no amount field and POSTs `ethAmount: 0` to `/api/coinbase/create-onramp-order`, which rejects it (`A valid ethAmount is required`). For non-US users who toggle to Coinbase, the hosted `CBOnramp` quote effect bails on `debouncedEthAmount <= 0` and leaves the Buy button permanently disabled. Net: only MoonPay can complete a generic top-up here — a regression from the previous `presetFiatAmount: 20` popup.

**Recommended fix:** either force MoonPay for the `wallet-fund` context, or thread `allowAmountInput` down to `CBOnramp`/`CBHeadlessOnramp` and add an amount input so a zero-amount order can't be submitted.

ui/pages/api/coinbase/create-onramp-order.ts
```diff
@@ -0,0 +1,258 @@
...
+      return res.status(400).json({
+        error: 'You must accept the Coinbase guest checkout terms to continue',
+      })
+    }
+    if (!partnerUserRef) {
+      return res.status(400).json({ error: 'partnerUserRef is required' })
+    }
+    if (!ethAmount || Number(ethAmount) <= 0) {
+      return res.status(400).json({ error: 'A valid ethAmount is required' })
+    }
```
pmoncada, line 141: This correct guard is what surfaces the wallet-Fund bug: the wallet Fund button (`PrivyConnectWallet`, `ethAmount={0}`) reaches `CBHeadlessOnramp`, which has no amount input and submits `0`, producing a 400 the user can't recover from. The validation is right — the caller needs to supply a non-zero amount (or let the user enter one) before this endpoint is hit.

ui/components/onramp/FundOnramp.tsx
```diff
@@ -0,0 +1,226 @@
...
+  )
+  // Track whether the user has manually switched so we don't override them.
+  const userSwitchedRef = useRef(false)
+  const [isUS, setIsUS] = useState<boolean | null>(null)
+
+  useEffect(() => {
+    // Only auto-select when the caller hasn't forced a defaultProvider.
+    if (defaultProvider) return
+    let cancelled = false
+    fetch('/api/coinbase/region')
```
pmoncada, line 76: Low priority: `FundOnramp` fetches `/api/coinbase/region` here, and the nested `CBOnramp` fetches the same endpoint again on mount, so the US Coinbase path hits this route twice per open. Consider detecting region once (e.g. lift to a shared hook/context) and passing `isUS` down to `CBOnramp` to avoid the duplicate request and a brief double loading state.

ui/components/moonpay/MoonPayOnrampModal.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
```
pmoncada: Low priority: with `FundOnrampModal` now the single entry point, is `MoonPayOnrampModal` (and the legacy `CBOnrampModal`) still referenced anywhere? If not, consider removing to reduce confusion about which modal is canonical.

pmoncada (commented):
Automated review of the on-ramp overhaul. Overall the architecture is solid (region-aware dual provider, race-guarded polling, redirect breadcrumb for Privy's no-return-URL onramp, and the Apple Pay domain-association conflict was resolved to the production file). One high-priority functional regression and a few low-priority cleanups are noted inline.

### 1. High — wallet "Fund" button is broken with `ethAmount={0}`
The generic wallet Fund action opens `FundOnrampModal` with `ethAmount={0}` and no amount input. For US users this defaults to the Coinbase Headless (Apple Pay) flow, which has no amount field and submits `ethAmount: 0`, which `create-onramp-order` hard-rejects (`A valid ethAmount is required`). For non-US users who switch to Coinbase, the hosted quote effect bails on `<= 0` and the Buy button stays permanently disabled (`allowAmountInput` is never threaded through here). Only MoonPay works for generic top-ups — a regression from the old `presetFiatAmount: 20` behavior.

**Recommended fix (pick one):**
- Route the generic wallet-fund context to MoonPay regardless of region (simplest), or
- Thread `allowAmountInput` through `FundOnrampModal` → `FundOnramp` for the `wallet-fund` context **and** add an amount input to `CBHeadlessOnramp` so a 0-amount order can never be submitted.

### Low-priority cleanups
- Double region fetch (see inline on `FundOnramp.tsx`).
- Possibly-dead `MoonPayOnrampModal.tsx` / `CBOnrampModal` now that `FundOnrampModal` exists.
- CSP still allows `pay.google.com` though web is now Apple-Pay-only.

(Not introduced by this PR: the malformed lock-icon SVG path in `CBOnramp.tsx` exists identically on `main`.)

pmoncada (approved):

pmoncada (commented):
Re-review after the latest changes. Thanks for the quick turnaround — the previously-flagged items are all resolved:

- ✅ **Wallet "Fund" with `ethAmount=0`**: `CBHeadlessOnramp` now has a USD amount input (spot-price → ETH, debounced, with a `> 0` guard + disabled button), `allowAmountInput` is threaded `FundOnrampModal → FundOnramp → CBOnramp → CBHeadlessOnramp`, and `PrivyConnectWallet` / `WalletInfoCard` set it. `useMoonPay` also takes a direct `address` now, avoiding the empty-`useWallets()` throw.
- ✅ **Duplicate region fetch**: `FundOnramp` passes `initialIsUS` to `CBOnramp` to skip the second `/api/coinbase/region` call.
- ✅ **Dead modal**: `MoonPayOnrampModal.tsx` removed.

The new `CreateTeam` onramp integration mirrors `CreateCitizen` (caching, auto-transaction, JWT resume, `shortfall * 1.15`); all referenced hooks/utils exist and signatures line up.

---

### 🔴 Merge blocker — production Apple Pay will break
`ui/public/.well-known/apple-developer-merchantid-domain-association.txt` currently decodes to `"domain":"moondao-git-coinbase-onramp-moondao.vercel.app"` (the branch preview), via commit `e90c7fda "chore(onramp): temporarily swap Apple Pay domain file to branch preview"`, and nothing reverts it. On `main` this file is for `www.moondao.com`.

If this merges as-is, the production `.well-known` file will advertise the Vercel preview domain, so Apple Pay merchant verification on `www.moondao.com` will fail — breaking the US Apple Pay onramp this PR adds.

**Recommended fix:** restore the production (`www.moondao.com`) domain-association file before merge. If preview-domain verification is needed for testing, register both domains with Apple and serve the production file on `main` (host the preview file via a preview-only path/env rather than committing it over the production file).

### Minor (non-blocking, carry-over)
`pay.google.com` is still allowed in `next.config.js` CSP (`script-src` / `connect-src` / `frame-src`) even though the web flow is now Apple-Pay-only (Google Pay was dropped). Safe to leave, but could be trimmed for a tighter CSP.

PR #1414, opened 2026-06-26
fix(teams): make Manage Team roles work and clarify admin access

mmoncada1:
## Summary

The Manage Team modal had two problems reported by users: managers "couldn't find any options" to assign a role, and trying to make someone an **Admin** had no working path.

Root causes (verified against `MoonDAOTeamCreator.sol`):
- The **Admin** option could never succeed — the team admin hat is a **1/1 hat minted to and worn by the team Safe**, so `mintHat(adminHat, person)` always reverts.
- The role dropdown rendered **blank while hat metadata loaded over IPFS**, so it looked empty.

The real elevated, human-assignable role is **Manager** (grants `isManager` / full administrative access). This PR makes that explicit and makes every action actually work per the on-chain permission model.

## Changes
- **Remove the unassignable Admin option**; clarify in the UI that Manager = full admin access.
- **Extract pure helpers** (`lib/hats/teamRoles.ts`): role detection, Safe-vs-direct routing, address validation, deterministic labels, and add/remove transaction builders.
- **Correct, verified routing**:
  - Add Manager / Remove Manager → **Safe** (admin of the manager hat).
  - Add Member → **direct `mintHat`** (connected manager is admin of the member hat).
  - Remove Member → **passthrough module** (`setHatWearerStatus` can only be called by the member hat's eligibility module).
- **Deterministic role labels** so the dropdown is never blank if IPFS is slow.
- **Modal UI redesign** (solid dark theme, role badges, inline remove, contextual Safe notices).
- **Dev-only preview page** at `/dev/manage-team` (returns 404 in production).
- **Tests**: unit + transaction-simulation that encode/decode the real Hats calldata for all four actions.

## Test plan
- [x] `yarn test:cypress-unit` — 25 `teamRoles` tests pass (incl. calldata simulation: target, selector, args, routing for add/remove member & manager). Pre-existing unrelated `citizenOnboardingImage` failures are not part of this PR.
- [x] Lint clean on all touched files.
- [ ] Manual: on a real/testnet team, connect a manager wallet and verify add-member (direct), add-manager (Safe queue), and removals.


pmoncada (approved):

pmoncada:
could you add a test that might be able to catch this issue and other issues like this in the future?

PR #1416, opened 2026-06-29
fix(teams): fix Manage Team modal layering and lock page scroll

mmoncada1:
## Summary
- Add `z-[9999]` to the Manage Team modal overlay so it paints above page content. The modal looked "see-through" because the custom overlay `className` had no z-index, letting page elements with positive z-indexes render on top of it.
- Make the inner panel `relative z-[10000]` (the previous inline `zIndex: 10000` did nothing since the element wasn't positioned) and add `shadow-2xl` for depth.
- Lock body scroll while the modal is open via a `useEffect` that sets `document.body.style.overflow = 'hidden'` and restores the previous value on close.

## Test plan
- [x] Open a team page and click **Manage Team**.
- [x] Confirm the modal is fully opaque and renders above all page content.
- [x] Confirm the page behind the modal cannot be scrolled while it's open.
- [x] Close the modal and confirm page scroll is restored.

### DIFF
--- a/ui/components/subscription/TeamManageMembers.tsx
+++ b/ui/components/subscription/TeamManageMembers.tsx
@@ -257,21 +257,28 @@ function TeamManageMembersModal({
     adminHatId
   )
 
+  useEffect(() => {
+    const previousOverflow = document.body.style.overflow
+    document.body.style.overflow = 'hidden'
+    return () => {
+      document.body.style.overflow = previousOverflow
+    }
+  }, [])
+
   return (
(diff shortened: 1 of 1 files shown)


pmoncada (approved):

ryand2d (commented):
Slightly pedantic, but terms are confusing, can we change:
"Current Members" > "Current Team"
and
"Add a Member" > "Add a Teammate" 

people seem to focus in on add member and can't find add manager

PR #1446, opened 2026-07-09
fix(dashboard): stop intermittent freeze — citizen redirect loop + RPC request storm

pmoncada:
## Problem

`/dashboard` intermittently loads for minutes and then crashes with "page unresponsive", across multiple machines. Adding a new Infura key did not help.

## Root cause

A self-sustaining feedback loop:

1. `CitizenProvider` resolves citizenship from a single SWR query to `/api/tableland/query` and **ignored the query's error state** — any failure (Tableland 429/5xx, our own API rate limiter, network blip) was treated as "not a citizen". It also overwrote the localStorage citizen cache with `undefined`.
2. `/dashboard` redirects non-citizens to `/`, and `/` redirects citizens to `/dashboard`. When citizen status flickered on a failed query, the user ping-ponged between the two pages.
3. Each bounce remounted the entire page tree (three.js globe included) and re-fired hundreds of RPC calls — dominated by the hats role-index scan (up to 200 tokens × 3 reads per contract) which also ran from the always-mounted nav dropdowns on **every** page.
4. That burst exhausted Infura (any key, client *and* server side) and our own rate limits, failing the next citizen query, restarting the loop. Console evidence: mass `Cannot read properties of undefined (reading 'buffer')` batch-decode failures, `getCitizenNFTByAddress` stack frames from the index page chunk while "on" the dashboard, React #421 hydration error, 500s from `/api/xp/citizen-referrals` (server-side quota exhaustion).

It recovers after a few minutes because rate-limit windows drain and a citizen query finally succeeds, warming the 5-minute cache — until the cache expires again. Hence the intermittency.

## Changes

- **`CitizenProvider`**: a failed citizen query is now "unknown", not "not a citizen". Keeps the last-known citizen, never poisons the cache on error, and stays in loading state (while SWR retries) if there's no known citizen — so route guards don't redirect on transient failures. Only a successful empty result clears citizenship.
- **`TopNavBar`**: Teams/Projects nav dropdown contents mount only while the dropdown is open (they were mounted, CSS-hidden, on every page), so their role-hat scans no longer fire on every page load. Scan results remain cached for 5 min, so opening the dropdown stays cheap.
- **`useRead`**: skips execution while any param is `undefined`/`null`. Previously `[terminal, undefined, token]` passed the truthy-array guard and viem threw `Cannot convert undefined to a BigInt` mid-encode (seen from `useTotalFunding` before `projectId` loads).
- **New `/api/rpc/[chainId]` proxy + client RPC rerouting** (`lib/rpc/chains.ts`, `lib/rpc/serverJsonRpc.ts`): browser RPC traffic goes through a Next API route that tries Infura first and falls back to public RPCs on 429/5xx/malformed responses, with its own IP rate limiting (100/s, 2000/min). Server/SSR keeps talking to Infura directly. This absorbs the `'buffer'` decode failures and transient 429s that degraded balances, quests, and check-ins.

## Not addressed here (follow-ups)

- thirdweb Insight 401s (`insight.thirdweb.com`) — the thirdweb API key's domain allowlist needs updating in the thirdweb dashboard (no code change).
- Moving the hats role-index scan server-side behind a cached route would further cut RPC volume.

## Test plan

- [ ] Signed-in citizen loads `/dashboard` with DevTools network throttling / simulated 429s from `/api/tableland/query` — stays on dashboard (spinner or last-known citizen), no redirect to `/`.
- [ ] Signed-in non-citizen visiting `/dashboard` still redirects to `/` once the query succeeds.
- [ ] Teams/Projects nav dropdowns populate correctly when opened; no hats-scan RPC traffic before opening.
- [ ] Dashboard featured-mission widget no longer logs `Cannot convert undefined to a BigInt`.
- [ ] `/api/rpc/1` (and other chains) proxies `eth_blockNumber` and falls back to public RPCs when Infura is rate-limited.

### DIFF
--- a/ui/components/layout/TopNavBar.tsx
+++ b/ui/components/layout/TopNavBar.tsx
@@ -183,10 +183,20 @@ const TopNavBar = ({
                         onMouseLeave={handleDropdownLeave}
                       >
                           <div className="min-w-56 max-w-xs w-full bg-gradient-to-br from-gray-900/98 via-blue-900/95 to-purple-900/90 backdrop-blur-xl border border-white/30 shadow-2xl py-2 px-2 rounded-xl">
+                          {/* Mount dynamic dropdowns only while open: their hooks
+                              (useTeamWearer / useProjectWearer) can trigger the
+                              on-chain role-hat index scan — hundreds of eth_calls —
+                              which must not run on every page load just because the
+                              nav bar exists. Results are cached (5 min TTL), so
+                              reopening is cheap. */}
                           {item.dynamicChildren === 'Teams' ? (
-                            <TeamsNavDropdown variant="desktop" />
+                            openDropdown === item.name ? (
+                              <TeamsNavDropdown variant="desktop" />
+                            ) : null
                           ) : item.dynamicChildren === 'Projects' ? (
-                            <ProjectsNavDropdown variant="desktop" />
+                            openDropdown === item.name ? (
+                              <ProjectsNavDropdown variant="desktop" />
+                            ) : null
                           ) : openDropdown === item.name ? (
                             item.children?.map((child: any, j: number) => {
                               if (!child.href) {
--- a/ui/lib/citizen/CitizenProvider.tsx
+++ b/ui/lib/citizen/CitizenProvider.tsx
@@ -235,6 +235,17 @@ export default function CitizenProvider({ selectedChain, children, mock = false 
       citizenRef.current = cachedData
       setCitizen(cachedData)
     } else {
+      // If the in-memory citizen belongs to a different wallet (wallet switch
+      // with no cache for the new address), drop it. Otherwise the data
+      // effect's error branch reads citizenRef and treats the new wallet as
+      // a citizen because the old wallet's NFT is still there.
+      const existing = citizenRef.current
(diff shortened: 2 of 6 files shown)


mmoncada1 (approved):

pmoncada:
## Follow-up: `/projects` freeze

Pushed an additional commit that targets the same Infura-saturating pattern on `/projects` (works on other pages; freezes like dashboard):

- **Shared IPFS cache** (`fetchProposalJsonCached`) — cards + author-address map no longer double-fetch every proposal CID
- **`useCitizens`** — one Tableland `owner IN (...)` query instead of N parallel `getOwnedToken` RPCs (chunked on-chain fallback)
- **`ProjectCard` visibility gate** — Engine/IPFS/NFT/hats work deferred until the card is near the viewport (`useOnceVisible`)

### Extra test plan
- [ ] `/projects` loads without going unresponsive with a long proposal list
- [ ] Scrolling loads card metadata (author, budget) as cards enter view
- [ ] Expanding a vote card still loads heavy data immediately
- [ ] Retroactive tab citizen filtering still works (Tableland batch path)

PR #1475, opened 2026-07-16
feat(proposals): let authors delete their own pending proposals

mmoncada1:
## Summary
- Adds a self-service **Delete Proposal** action so authors can remove their own proposals (duplicates, withdrawn, mis-submitted) instead of an admin blocking them by hand.
- Deleting **reuses the existing edit flow** — no new on-chain status, no new API endpoint. It re-pins the proposal JSON with a `deleted: true` flag and updates the row's `proposalIPFS` via the existing `/api/proposals/submit` UPDATE path.
- The frontend hides any proposal whose JSON carries `deleted`, so it disappears **for everyone**.

### Frontend
- `components/nance/ProposalEditSection.tsx`: **Delete Proposal** button + confirmation modal, reusing the existing author-only / not-Approved gate. On success, redirects to `/projects`.
- `pages/projects/index.tsx` and `pages/governance-proposals.tsx`: skip proposals whose JSON has `deleted` (these already fetch the proposal JSON).
- `pages/project/[tokenId].tsx`: returns `notFound` when the proposal JSON has `deleted`, so a direct link can't resurface it.

## Notes
- Author-gating and write auth are inherited from the existing edit flow (author-only UI; `/api/proposals/submit` verifies the Privy session owns the wallet).
- Soft, reversible, and cheap: the on-chain row is untouched apart from its `proposalIPFS` pointer.

## Test plan
- [ ] As the proposal author, click Delete → confirm → redirect to `/projects`.
- [ ] Confirm the proposal disappears from `/projects` and `/governance-proposals` after revalidation.
- [ ] Confirm `/project/<mdp>` now returns 404.
- [ ] Confirm the Delete button is hidden for non-authors and for Approved proposals.


pmoncada (approved):

pmoncada:
Random test is failing, probably unrelated, can you fix in a diff pr?

PR #1553, opened 2026-08-28
fix(lunar-atlas): apply corrections surfaced by Deloitte's lunar economy report

mmoncada1:
## Summary

Reviewing Deloitte's *Building the Lunar Economy* report against our own dataset turned up several races priced on facts that have since moved. As written, they would have resolved on criteria NASA has abandoned, or against a field that no longer exists.

**Fission power — wrong threshold.** The race resolved at 40 kWe, the figure from the 2022 concept studies. NASA's directive raised it to a **100 kWe minimum** on HALEU fuel with a 15 t lander allowance, and the NASA–DOE memorandum set launch readiness at Q1 FY2030. Criteria, window and sources updated.

**Rover — wrong field.** The race still showed a three-way tie led by Moon RACER at 42%. NASA issued Phase 1 task orders to only two vehicles on 26 May 2026: Astrolab's **CLV-1** ($219M) and Lunar Outpost's **Pegasus** ($220M), both due on the surface by **2028**. The projects now carry the awarded names, the award values as milestone metrics, and 2028 dates. The race also had *no criteria at all*, so it gained four taken from NASA's revised requirement (crewed + autonomous + teleoperated, ~200 km at ~10 km/h, one year of service, landed before Artemis IV).

**Naming.** Astrobotic → Voyager Lunar Systems (Voyager Technologies acquisition, July 2026). Nokia Bell Labs → Modul8 (spun out via Celestial, 2026).

**Moon Base I/II/III.** Added as milestones on the landers that fly them — Blue Moon MK1, Griffin-1 to Nobile Crater, Nova-C Trinity to Reiner Gamma. These are the nearest-term resolvable events on the board and were missing entirely. Note that these three landers are not competitors in any race, so they surface in project panels and the timeline rather than on the base itself (see follow-ups).

**Comms.** Added a bandwidth criterion. Coverage and interoperability decide who can sell a service; throughput decides whether it is worth buying.

## Note on Moon RACER

It is not deleted. Two constraints pushed against that: a selector test asserts every project eventually stands on the finished base, and dropping it from the race roster would remove a plot from the rover district, since the surface renders a race's declared competitors and nothing else. It sits at 0.04 as a tail on a later phase of a multi-phase programme worth more than $4B, with its summary stating plainly that it lost Phase 1. Happy to cut it properly in a follow-up if preferred.

## Scope

Data only. No project ids, locations or footprints change, and every race roster keeps its competitor count, so district geometry and the 3D model bindings are untouched.

## Test plan

- [x] `yarn test:cypress-unit` — 555 passing. The 9 `citizenOnboardingImage` failures (`ReferenceError: File is not defined`) reproduce identically on a clean tree and are unrelated.
- [x] `npx tsc --noEmit` — no new errors; the only 4 are pre-existing `import.meta` issues in `scripts/verify-deprize-moonbase-sepolia.ts`.
- [x] Dataset integrity checked: valid JSON, unique milestone ids, every goal/project cross-reference resolves, no implied odds on non-competitors.
- [ ] Spot-check the rover and power districts in the moonbase view to confirm the panels read correctly.

## Follow-ups not in this PR

- **Two geometry rosters under-count their district.** `ROSTERS` in `lunar-atlas-baseplan.cy.ts` is mirrored by hand, and two entries were never updated when competitors were added: `isru_plant` lists 3 plots against 4 real competitors (Cislune missing) and `comms_pnt` lists 4 against 5 (Solstar missing). The other six districts are correct. This is the same failure the construction roster's own comment warns about — "at four it silently stopped covering that branch, which is how a fifth lot came to be placed off the end of its own avenue" — so those two districts are currently packed untested at their real size.
- **Nine competitors still render a generic model**, five of them the entire construction race (ICON, Redwire, Astroport, AI SpaceFactory, Voyager), plus Modul8 (the comms front-runner), Solstar, Cislune and Westinghouse's eVinci.
- **No cargo-lander race exists.** The five CLPS-class landers sit in the dataset with no race, so they never appear on the base — including the three Moon Base missions added here, which are the nearest-term resolvable events on the whole board.
- A "survive the night" radioisotope race (Zeno et al.) would need new competitors in the power race, which changes that district's plot count.

### DIFF
--- a/ui/lib/lunar-atlas/seed/atlas.dataset.json
+++ b/ui/lib/lunar-atlas/seed/atlas.dataset.json
@@ -201,17 +201,22 @@
     },
     {
       "id": "astrobotic",
-      "name": "Astrobotic Technology",
+      "name": "Voyager Lunar Systems",
       "kind": "company",
       "country": "USA",
       "website": "https://www.astrobotic.com/lunar-delivery/",
       "brandColor": "#38BDF8",
-      "summary": "Astrobotic builds the Peregrine and Griffin lunar landers plus CubeRover mobility, delivering payloads to the surface under NASA's CLPS program.",
+      "summary": "The Pittsburgh lander company founded as Astrobotic Technology, renamed Voyager Lunar Systems when Voyager Technologies closed its acquisition in July 2026 with founder John Thornton still leading it. Builds the Peregrine and Griffin landers plus CubeRover mobility, delivering payloads under NASA's CLPS program and the Moon Base missions.",
       "sources": [
         {
           "label": "Astrobotic — Lunar Delivery",
           "url": "https://www.astrobotic.com/lunar-delivery/",
           "retrievedAt": "2026-07-27"
+        },
+        {
+          "label": "SpaceNews — Voyager completes acquisition of Astrobotic",
+          "url": "https://spacenews.com/voyager-completes-acquisition-of-astrobotic/",
+          "retrievedAt": "2026-08-28"
         }
       ]
     },
@@ -313,17 +318,27 @@
     },
     {
       "id": "nokia",
-      "name": "Nokia Bell Labs",
+      "name": "Modul8",
       "kind": "company",
-      "country": "Finland",
-      "website": "https://www.nokia.com/about-us/newsroom/articles/nokia-bell-labs-lunar-network/",
+      "country": "USA",
+      "website": "https://www.modul8space.com/missions/",
       "brandColor": "#124191",
-      "summary": "Nokia Bell Labs adapted terrestrial LTE/4G for the lunar surface under a NASA Tipping Point award and flew the first cellular network to the Moon aboard Intuitive Machines' IM-2 in 2025.",
+      "summary": "The Space Communication Solutions venture built inside Nokia Bell Labs, rebranded Modul8 and spun out into an independent company through Celestial Acquisition Corp. in 2026, with Nokia staying on as a major shareholder. It adapted terrestrial LTE/4G for the lunar surface under a NASA Tipping Point award and flew the first cellular network to the Moon aboard Intuitive Machines' IM-2 in March 2025; it is now integrating that system into Axiom Space's AxEMU spacesuit.",
       "sources": [
         {
           "label": "Nokia — Lunar network",
           "url": "https://www.nokia.com/about-us/newsroom/articles/nokia-bell-labs-lunar-network/",
           "retrievedAt": "2026-07-27"
+        },
+        {
+          "label": "Nokia — A proposed spin-out to launch Modul8 as an independent company",
+          "url": "https://www.nokia.com/blog/an-important-step-for-nokias-modul8-a-proposed-spin-out-to-launch-an-independent-company/",
+          "retrievedAt": "2026-08-28"
+        },
+        {
+          "label": "Modul8 — Missions",
+          "url": "https://www.modul8space.com/missions/",
+          "retrievedAt": "2026-08-28"
         }
       ]
     },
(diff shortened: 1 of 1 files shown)


mmoncada1:
## Follow-up: the two under-counted rosters are hiding a live geometry bug

Correcting `ROSTERS` in `lunar-atlas-baseplan.cy.ts` to the real competitor counts makes two districts fail `leaves the inward corner lots clear of the perimeter road`. Measured against the 74.3 m clearance floor (ring road at `RING_RADIUS_M` 67, `ROAD_HALF_M` 6.3, plus 1 m):

| District | Project | Footprint radius | Inner edge | Overlap |
|---|---|---|---|---|
| `isru_plant` | Cislune PERDEX / CISORT | 9.5 m | 68.9 m | 5.4 m into the road |
| `comms_pnt` | Solstar Lunar WiFi | 7.5 m | 72.9 m | 1.4 m into the road |

Both stand on the pavement today; the omitted roster entries are the only reason CI is green.

**Why those two.** The crossroads case fills corners diagonally (`CORNERS`), so the second-largest plot always takes an *inward* corner, where the inner edge is `MAIN_LOOP_M - 8.1 - 2r` = `87.9 - 2r`. Any plot with `r >= 6.8` overlaps. Cislune and Solstar are oversized for exactly one reason: neither has a bespoke model, so both inherit the generic footprint (`TYPE_SIZE_M.isru_plant` 19 → 9.5, `comms_pnt` 15 → 7.5). The model gap and the geometry bug are the same gap.

This also blocks adding Queqiao-2 to the comms race, which is otherwise a strong candidate — it has been operational since 2024, relayed Chang'e-4 and Chang'e-6, and does 1–500 Mb/s, so it clears the new bandwidth criterion that arguably none of the current five do. With no model it would come in at the generic 7.5 m, making three such lots in a district that already has one on the road.

**Three ways out, deliberately not chosen here:**

1. Give Cislune, Solstar and Queqiao-2 real footprints via `PROJECT_SIZE_M`. Most accurate — the generic sizes are placeholders that badly overstate small hardware, and it matches ESA (1.04), Crescent (1.52) and IM (1.13). Cost: they render the generic model shrunk until each gets its own.
2. Assign corners by size so large plots take outward corners. Fixes the whole class permanently, but works against the documented reason for diagonal fill (a two-competitor race should read as a junction).
3. Raise `MAIN_LOOP_M` from 96 to ~102 (needs > 101.4 to clear a 9.5 m inward lot). One constant, but moves every district outward.

Not addressed in this PR, which is data-only.

mmoncada1:
**Update on the geometry follow-up above — it's fixed, in #1554.**

The earlier note described the Cislune/Solstar perimeter-road overlap as an
open bug and listed three possible fixes. It has since been fixed, so that
comment should be read as history rather than as outstanding work on this PR.

The chosen fix was the third framing of option 2: rather than giving the two
projects smaller footprints or pushing `MAIN_LOOP_M` outward, `districtSlots`
now seats a district's biggest plots on the OUTWARD corners and its smallest
on the inward ones. The fill order is unchanged, so a two-competitor race
still gets one lot either side of both streets.

Confirmed margins to the perimeter road:

| District | Plot | Before | After |
|---|---|---|---|
| `isru_plant` | Cislune (9.5 m) | −5.4 m | +29.8 m |
| `comms_pnt` | Solstar (7.5 m) | −1.4 m | +29.8 m |

`ROSTERS` in `lunar-atlas-baseplan.cy.ts` is also corrected to its true counts
there — the under-count was the reason neither overlap ever failed a test.

**Nothing in this PR changes.** #1554 is stacked on this branch and will
retarget to `main` on merge; this one is still green and waiting only on
review.

pmoncada (approved):

PR #1574, opened 2026-09-11
Light every regolith surface with the real reflectance law, not a baked hillshade

mmoncada1:
The ground was a **picture of shading**, baked once by a Python script and pasted on. It is now lit for real, per pixel, by the same light as everything else in the scene.

## Why this matters

Lunar regolith does not behave like a generic surface. It gets *brighter* when the sun is behind you (opposition surge) and barely dims at grazing angles. The Lambertian shorthand every engine reaches for is about **4x wrong** at this scene's phase angles, and that error had already been absorbed three separate times in this codebase — in the environment map, in the shadow fill, and in the graded road surfaces. There is now one function to call instead, and a comment saying so.

The old bake also flattened the thing that makes this subject legible: a sunlit white panel sat **0.59 stops** above the painted ground. It now sits **4.6 stops** above the computed one. Six tenths of a stop of separation between white metal and lunar soil is most of why this scene kept getting described as grey plastic.

## What is in here

- Terrain shaded by the regolith BRDF per pixel, replacing the baked hillshade.
- The separate render pass that existed only to catch shadows on the terrain is **deleted** — the terrain shades itself now.
- Roads, graded pads, spoil heaps and loose rubble moved onto that same law. They each had their own approximation before, so a road crossing the ground did not quite match the ground.
- The fake overhead sky-glow is gone. There is no atmosphere, so nothing fills a shadow except light bounced off nearby regolith, which arrives from **below**, warm-grey, at about 4% of the sun. A hemisphere light says the opposite and it is what made every shadowed face read as a plastic toy under a softbox. That bounce is now carried by the environment map, which gets it directionally right for free.
- Ground stains became multiplicative albedo factors rather than absolute painted values, so they are exposure-invariant by construction.
- A procedural detail tile supplies relief below the DEM's 5 m resolution.

## Payload

**-6.5 MB.** `albedo.jpg` is deleted; the ground's colour is computed rather than shipped. `build-southpole-assets.py` shrank by about two thirds since it no longer bakes that texture.

## Reviewing this

The physics lives in `lib/lunar-atlas/regolith.ts` and the GLSL surgery in `lib/lunar-atlas/regolithShader.ts`. Both are covered by unit tests — **233 passing** — including the shader patches, which are asserted against three's real shader source so a missed anchor fails a test rather than blanking a surface at mount.

`SouthPoleTerrain.tsx` is net *smaller* despite doing more, because the bake machinery and the shadow-catcher pass both went away.

## Known consequence, deliberately not addressed here

The ground is about **4.1 stops darker** than the bake it replaced. That is correct — sunlit regolith at 0.12 albedo simply is that dark next to hardware — but it is a large change from what people were used to. Re-anchoring the exposure is handled in a later PR in this stack, where a moving sun turns it into a derivation rather than a judgement call.

## Stack

1. **this PR**
2. Add the lunar sun path and skyline field
3. Add the sun scrubber and true-sun mode


pmoncada (approved):
resolve merge conflict

PR #1592, opened 2026-09-17
refactor(nav): collapse the navbar from 9 groups to 5

mmoncada1:
## What this does

The navbar presented **9 top-level groups and 41 links**. This reduces it to **5 groups and 23 links**, and moves per-user entries behind the avatar.

No URLs change — only the labels and grouping that lead to them — so nothing needs a redirect.

| | Before | After |
| --- | --- | --- |
| Top-level groups | 9 | 5 |
| Static links in the menu | 41 | 23 |
| Destinations reachable 2+ ways | 8 | 0 |
| Hand-maintained copies of the tree | 3 | 1 |

Full write-up in `docs/NAVIGATION_SIMPLIFICATION_PLAN.md`.

## Why

Four things were wrong, and they compounded:

**Six landing pages were effectively unreachable.** Any item with a dropdown rendered as a `<button>` whose tooltip read, verbatim, *"Single click: open menu. Double click: go to page."* Nobody double-clicks a nav item.

**Campaigns outranked the product.** Moon Base Zero, DePrize and Send Frank to Space each held a top-level slot — a third of the bar — at the same weight as Citizens, Teams and Projects.

**Eight destinations had more than one front door.** `/contributions` was linked four times under two different names; `/projects`, `/mooney`, `/launch` and `/network?tab=teams` were each linked three times. Deduplicating removed 15 links without removing a single destination.

**The tree existed in three hand-maintained copies** — `useNavigation`, `ExpandedFooter` and the footer's own arrays — which had already drifted. "Create a Team" pointed at `/team` in the bar and `/join` in the footer, and the footer never grew columns for Moon Base Zero or DePrize.

The bar had also outgrown its layout. `TopNavBar` carried comments recording that the row *"overflowed its own box at every window size, including a 2560px one"* and that eight icons *"cost 192px"* and were hidden below a custom `navicons` breakpoint. That breakpoint is deleted here; five items fit with the icons showing.

## New structure

**Network** → `/network` · Citizens, Teams, Map, Jobs, Marketplace
**Projects** → `/projects-overview` · Explore Projects, Propose a Project, Submit a Contribution
**$MOONEY** → `/mooney` · Get, Lock & vote, Bridge
**Missions** → `/launch` · Send Frank to Space, Moon Base Zero, DePrize
**Learn** → `/info` · Documentation, News & Updates, Governance, Roadmap

Behind the avatar: Your Profile, Your Teams, Your Projects & Proposals, Create a Team, Submit a Contribution. Signed out, that slot is a single **Join** button, sized to match the Sign in button beside it.

Projects lands on the overview rather than the list, because someone clicking it cold usually wants to know what the project system *is* before they want forty rows of it. Governance sits under Learn, since reading how the DAO decides things is a different errand from acquiring or locking the token — Governance Proposals, where voting happens, stays with $MOONEY in the footer.

### Town Hall is unlinked

Not demoted to the footer — removed from the bar, the footer, and the search index.

`lib/townhall/summaries.ts` reads from ConvertKit, and `getTownHallBroadcasts` catches every API failure and returns `[]`. The page then renders *"No town hall summaries available yet. Check back soon!"* — so an expired key and a quiet week are indistinguishable, which is why this went unnoticed for months. The local `CONVERT_KIT_V4_API_KEY` currently returns **401 invalid**.

The frontend is not the bug. Relink it once the feed is publishing again; separately, that `catch` should surface an error state rather than an empty one.

## Key changes

- **`lib/navigation/nav-config.ts`** (new) — one source of truth. The bar, drawer and footer all read from it. A group's `footerOnly` list is how demoted reference pages (Proposal Template, Press, Resources, Governance Proposals, Treasury, FAQ) stay reachable without occupying the bar.
- **`TopNavBar`** — group label is a `NavLink`, chevron is a separate `<button>` with `aria-expanded`. Double-click behaviour and its tooltip are gone.
- **`Sidebar/NavigationLink`** — same split on mobile. Previously the whole row was a `Disclosure.Button` that pushed a route from an inner div's `onClick` *while* toggling the panel.
- **`account/AccountMenu`** + **`account/MobileAccountSection`** (new) — per-user entries. This also keeps `useTeamWearer` / `useProjectWearer` (on-chain role-hat index scans) out of the public nav entirely.
- **`Sidebar/TeamsNavDropdown` → `account/UserTeamsList`**, same for Projects — stripped of their public links, which now live in the Network and Projects groups.
- **`ExpandedFooter`** — columns generated from `NAV_GROUPS` instead of six hand-written arrays.

## Screenshots

Five items with icons, and the Missions group holding all three campaigns. (Screenshots captured locally at 1800px via Playwright — drag them into this thread to attach: `/tmp/nav-closed.png`, `/tmp/nav-missions-open.png`.)

## Testing

- `tsc --noEmit` — no new errors. The pre-existing failures in `lib/safe/useSafe.tsx` and `scripts/` are untouched and also fail on `main`.
- No lint errors in any changed file.
- Dev server: `/`, `/projects`, `/mooney`, `/launch`, `/info`, `/jobs`, `/network` all 200. `/moonbase` and `/deprize` 307 to `/gate`, which is the pre-existing access gate and not related to this change.
- Verified in the rendered HTML that the removed duplicate labels are gone and the footer is 5 config-driven columns.

## Not in this PR

- **`GlobalSearch` still hardcodes its own 33-entry index.** It is unaffected — no URLs changed — but it is the third copy of the tree and should derive from `nav-config` too. Search also becomes more load-bearing now that the long tail is out of the bar, so giving it a visible affordance and a keyboard shortcut is the natural follow-up (step 5 in the plan).
- **The demotions are not validated against traffic.** The audit was structural; no analytics were available. Worth confirming Press, Resources, Proposal Template and Projects Overview really are low-traffic before this merges.

## Open question

The signed-in landing page still points at `/`. There is a dedicated `/dashboard` page — should the logo and Join flow target that instead?


mmoncada1:
### Town Hall restored — the premise for removing it no longer holds

This branch dropped Town Hall from the bar, the footer and the search index, and encoded the reasoning in three comments: the page had been showing "No town hall summaries available yet" for months.

That diagnosis was wrong. The page was never broken — its **publishing pipeline** had been failing silently since May, so the page was faithfully rendering an archive that had stopped growing. [#1603](https://github.com/Official-MoonDao/MoonDAO/pull/1603) fixes that and backfills the gap; the archive is now current at 114 summaries.

Those comments named their own trigger for relinking — *"once the feed is fixed"* — and that has happened. Shipping them as written would leave a working page unreachable from nav, footer and search, plus three notes telling the next engineer a fixed thing is broken.

So Town Hall is back in the **Learn** dropdown, the search entry is restored, and the three stale comments are gone. Learn goes to five children; the bar to 24 links. The planning doc is corrected to record what actually happened rather than the original misdiagnosis.

Verified `/townhall` activates Learn and only Learn via `isGroupActive`, and that the footer column picks it up from `NAV_GROUPS` automatically.

Nothing else about the restructure changed.

**Suggested ordering:** merge #1603 first, though these are independent — the archive is already live, since the backfill wrote to ConvertKit rather than to code.

pmoncada (approved):

PR #1603, opened 2026-09-17
fix(townhall): restore the summarizer — owner-authenticated captions + live GROQ model

mmoncada1:
## Three separate breakages, not one

The Town Hall page hadn't had a new summary since **May 4, 2026**. The working theory was expiring YouTube cookies. It was none of that — and it was three faults stacked, each hidden behind the one before it.

**1. Captions were fetched anonymously.** YouTube does not answer anonymous caption requests from datacenter IPs. Same video, same code, two networks:

| Video | In GitHub Actions | From a laptop |
|---|---|---|
| `f_jXiCdzIxQ` | "No caption track available" | 51,267 chars |
| `ELcZETA5yzw` | "No caption track available" | 54,127 chars |

The workflow pinned `TRANSCRIPT_SOURCE: captions`, so the one strategy that cannot work on a runner was also the only one permitted. Every scheduled run since May failed in under 90 seconds.

**2. The GROQ model was decommissioned.** Only visible once transcripts started arriving: GROQ retired the whole Llama family, so `llama-3.3-70b-versatile` returns 404 `model_not_found`. Summarization would have failed even with a perfect transcript.

**3. Nothing alerted.** A stale `/townhall` page is indistinguishable from a quiet one, so it rotted for four months. The Vercel cron was removed on **May 5** (`2d65047c`) in favour of GitHub Actions — one day after the last good summary, which is exactly where the gap starts.

## The fix

MoonDAO owns these videos, so stop scraping them. New `owner` transcript source downloads captions through the YouTube Data API authenticated as the channel, ahead of the public endpoint and the yt-dlp audio fallback. No cookies, no PO tokens, no dependence on runner IP.

- `TRANSCRIPT_SOURCE: auto` so one dead strategy degrades instead of failing the run.
- `openai/gpt-oss-120b` replaces the retired model (same 128k context, so chunking is unchanged). The name was hardcoded in four places; three now read `DEFAULT_MODELS.llm`.
- Discord alert on workflow failure.
- `yarn oauth:setup` mints the refresh token and reports the consenting account **and** the channel it controls.
- `yarn captions:check <id>` reports every step separately, so failures identify themselves.
- README corrected — it still documented the Vercel cron removed in May.

> **Why credentials kept "expiring":** Google expires refresh tokens after **7 days** while the OAuth consent screen is in **Testing**. Setting it to **In production** stops that. A console setting, not a code problem — which is why previous workarounds never held. A Brand Account **Manager** can complete the flow; you needn't be the owner.

## Verified in production

```
1. OAuth credentials      ok
2. Token exchange         ok
3. Authenticated channel  MoonDAO (UCoyGiF4NdGV9YfKFEpm3Hvw) — matches configured
4. f_jXiCdzIxQ            ok — asr/en, 102,413 bytes SRT -> 50,751 chars
4. ELcZETA5yzw            ok — asr/en, 108,318 bytes SRT -> 53,460 chars
2/2 video(s) produced a transcript through the owner API.
```

This settles the one risk I couldn't pre-test: **YouTube does serve ASR tracks to the channel**, which matters because town halls are live streams and ASR is all they have.

Three town halls were then processed end to end and **are now live on https://moondao.com/townhall** — the first new summaries in four months.

Two further things found and fixed by actually running it:
- **YouTube Data API v3 was not enabled** on `moondao-townhall`. Now enabled.
- The first published summaries contained **broken URLs** — ASR renders "MoonDAO" as "Mundow"/"Moonow" inside spoken links, and the correction list missed those variants, shipping "mundow.com/frank" to readers. Added the variants plus a domain-normalising rule; the affected broadcasts were deleted and regenerated clean.

## Remaining

- [ ] Backfill the rest (~19 more). `yarn missing --days=180` lists 29 candidates, but roughly seven are short promo clips rather than town halls and should be skipped.
- [ ] Review and send the drafts in ConvertKit — they are created unsent by design.

## Note on #1592

That PR unlinks Town Hall from the nav as "broken". It wasn't broken, just four months stale, and it is now current. The nav link should stay.


pmoncada (approved):

mmoncada1:
### Verified in GitHub Actions, not just locally

The whole failure mode here was "works on a laptop, fails on a runner", so local verification wasn't sufficient. Triggered the workflow on this branch against `lj7X2ODjbKc`, a real unprocessed town hall:

```
Downloading asr caption track (en) for lj7X2ODjbKc as channel owner...
Transcript ready: 57207 characters (via owner API)
Using owner-API transcript (asr track, en, 57207 chars).
Summary generated successfully
Broadcast draft created successfully: 25972391
Successfully processed: 1
Failed: 0
```

[Run 35297611549](https://github.com/Official-MoonDao/MoonDAO/actions/runs/35297611549) — **the first success for this workflow since May**, in the same environment that had failed every Friday for four months. The runner pulled captions that the previous anonymous path could not see from that IP, confirming the owner route is what fixes it.

mmoncada1:
### Backfill complete — 114 summaries live

Ran the backfill. Filtered the 29 candidates by **duration** rather than title, which separated them cleanly: the 7 skipped are all 1–2 minute announcement clips, the 18 kept run 29–148 minutes.

**17/18 processed.** The one miss is `7uYEKyEtnZw` — today's town hall, whose auto-captions YouTube hasn't generated yet. That's the correct "not ready" path rather than a failure, and Friday's scheduled run will pick it up.

The archive is now **114 summaries with zero broken URLs**. One older broadcast (April 9) still contained "Mundow" from before the spelling fix; it was regenerated too.

### One more bug this exposed

Backfilling five months in a single run made a latent display bug obvious: the list **sorted** on the video date embedded in each broadcast but **displayed** `published_at`, the date the summary was generated. With summaries normally produced the morning after each town hall those were nearly identical — but after a bulk backfill, every town hall from April onward showed the same September date. The April 9 episode was labelled September 18.

Fixed in `ui/lib/townhall/summaries.ts` by displaying the same video date the sort already uses. All 114 broadcasts carry the marker, so nothing hits the fallback.

